Issue #457, opened 2014-09-30
Support Operator framework 


prudhomm:
We should support an Operator framework in order to handle complex operators and preconditioners.On top of that we should define a language to support these operations. The language could be defined using boost::proto.

For example 
```c++
// we want to solve S x = b
auto S = block() << A << B << trans(B) << zero;
// we define a preconditioner for S
auto prec  = block() << inv(A,"invA") << B << trans(B) << trans(B)*inv(A,"invA")*B;
backend()->solve( _matrix=S, _prec=prec, _solution=x, _rhs=b );
````
@mszopos @vincentchabannes what do you think ?

the string `"invA"` provides a name for a backend to use

PR #458, opened 2014-09-30
Operator Framework (#457)

prudhomm:
- define some operators
 - define a language to work with blocks
 - allow this to create preconditioners
 - propose an implementation of BTPCD so that we can control exactly what is going on #460

this pull request implements #457 @vincentchabannes @mszopos


prudhomm:
@jbwahl @vhuber you are also concerned by this branch: the preconditioner that is being developed can be extended to aerothermal flows

vhuber:
The `OperatorMatrix::applyInverse` function does not allow to pass a preconditioner to the backend. Do I have to make it possible or there is a reason to not make it possible ? (the AFP pc need a specific pc inside).

Issue #542, opened 2015-03-11
Ginac : Using the same filename doesn't update the expression


aancel:
I recently encountered an unexpected behavior, while using Ginac expressions.
Let's say that I have the following code:

``` c++
void MyFunction(std::string expression)
{
...
auto f = expr( expression, "f")
...
}
```

This function is called several times to update the data produced by an exporter.
The first time it is called, eveything is ok, the expression is updated and the resulting data are ok.
However, when calling this function again with a different expression, the following output data remained the same, as with the first call. We know that this problem can be solved by not providing a filename, but we didn't want to multiply the of .so files generated by our app (as it is an interactive process).

We had a look in the code with @vhuber, and noticed that, when using the same filename, references to previously compiled expression were not closed properly on the ginac side and not updated in Feel++ data structures.

We made a fix for this issue, I'll issue a PR for our proposition.

PR #543, opened 2015-03-11
Issue #542:

aancel:
Unlink ginac files and update Feel++ structures when the user uses
the same file name on successive expr( ... ) calls,
to ensure that the expression is updated


vincentchabannes:
Warning, there are several use of ginac. Please add also your problem in test_ginac.
I don't understand why you rename the key expression. It's probably break codes which try to search if an expression has been already compiled (the filename is not already known). Also if I remember, the filename can be empty. And are you sure that the unlink is apply only in your specific case.

If you specify the filename, then you have access to the desc file in addition to the ginac files which describe the expression. If the desc file exist, you can load the expression inside and get the key for the manager. And if the expression in this desc file is not the same that the current expression, then apply the unlink and the erase the key in manager.

vhuber:
@vincentchabannes The key expression is renamed thanks to the filename - that can't be empty since it is automatically generated (in the `ginacex.hpp` file, line 106) if the user does not provide one - to assert no expression **with the same filename** has been generated (in the current run).
About the unlinking process, GiNaC search for a linked dynamic library created from an expression with the key `filename".so"` to unlink and does nothing if it is not found.

I guess you have misunderstood the point here:
```c++
void toto(std::string s){
auto f = expr(s,"file");
}
toto("2*x+4:x:y");
toto("sin(x)*cos(x):x:y");
```

The key point here is to load two dynamics libraries that have the same filename.

When parsing the first string, the GinacExpressionManager will reference the expression and inside GiNaC, the created dynamic library - named "file.so" will be loaded using `dlsym`.
At the second string, inside GiNaC, we will load "file.so" **that have actually changed** but as long as there is no **dlclose** to answer the first `dlsym`, the same handle as the first one will be provided.
Thus, we have to unlink the first one. That force us to remove the old entry in the GinacExprManager.

vincentchabannes:
I have understood your problem and indeed I have not take into account this when I done the mangement of ginac lib. (but the thing not clear for me is why you don't write all ginac file. It's really very small file.)
About your solution, changing the key name will be problematic for me, I think (I need to check). And the line if(key.size() >=.... ) is quite dangerous. I have proposed another solution. I have no time today and the next days, we can discuss at the FUD. But commit this if you are sure that all case are taken into account.

aancel:
@vincentchabannes Did you have the time to check the proposed modifications ?
How do you think the `if( ... )` line should be changed ? I tried to make the check as bulletproof as I could

vincentchabannes:
Not really! and I have no time at this moment!
I repeat : change the name of the key can be problematic (It change a main idea in the ginac management).
I think we can support this feature without to change the name of the key. Probably I will be in Strasbourg the next week, we can talk about this (I forgot at the FUD). Also I need to understand your context. How your expression change at each successive expr()? if you want to do this, it's not just to change a parameter in the expression like the time, no?
And don't forget that build ginac lib at runtime can be problematic on supercomputer.

aancel:
@vincentchabannes Ok, no problem. we can discuss this when you have more time.
I guess it is not mandatory to change the key. If I remember well, we changed the key name essentially to record the name of the file used to store the code and the dynamic library. It was done to ensure that the expression is correctly rebuilt when you use several times a declaration like:
`auto f = expr( "...", "f" );

If called several times, this expression will be ok the first time, but it won't be refreshed on the following calls. It might happen when a user has a function called several times with a code similar to the previous one in it. I know that the your can remove the expression name, but I think that doing the changes here will enable to have a code working the way a user expects it to work, i.e. if you use the same expression name, the previous expressing is replaced and not available anymore.

vincentchabannes:
yes I have understood and we need to support it. I asked you a concrete exemple.
Generally, the users must not specify the filename. It's automatically done by feel. And thus, this issue not appear. Specify the filename is only usefull for store the ginac lib generated in a specific path. For example on a supercomputer, I run on frontal or one node a programm which generate my expressions and then I launch my simulation on many nodes which use this files. It avoid to have some troubles with i/o.

aancel:
The only concrete example I would see is the one you are mentionning, where you want to use only a specific file.
This problem arised in the ParaView plugin we developed with @jbwahl. We are re-entering a function that reconfigures the simulation and needs to regenerate the expression with a string given through the ParaView interface. When I first worked on this, I just copied-pasted the qs_laplacian example where the expression name is defined and end up with the error previously described.

prudhomm:
this has become irrelevant. thanks anyway for this work

Issue #587, opened 2015-05-19
Optimization of geometric transformation for MultiScaleImage


prudhomm:
@LANTZT the point coordinates for integration are obtained from the fine grid with respect to the coarse one and the geometric transformation. Since the pixelsize is always the same, when going from super-pixel to another, the point coordinates are just translated. This optimization can be coded into the geometric transformation to speed-up the assembly process.

Issue #586, opened 2015-05-19
Add examples for MultiScaleImage


prudhomm:
@LANTZT could you add some examples for multiscale images.

```cpp
// given L the coarsening level of the  mesh with respect to the fine mesh
// define f and g images on the fine grid
auto Xhc = ...;
auto u = Xhc->element(), u = Xhc->element();
auto l = form1(_test=Xhc );
l = integrate( _range=elements(mesh), _quad=IMImage(L,pixelsize), _expr=msi(f)*id(v));
auto a = form2(_test=Xhc, _trial=Xhc);
a = integrate( _range=elements(mesh), _quad=IMImage(L,pixelsize), _expr=idt(u)*id(v));
a.solve( _rhs=l, _solution=u );
// u is the L2 projection of f on Xhc 
```

vhuber:
@prudhomm @LANTZT The associated test fails.
See [cdash](http://my.cdash.org/testDetails.php?test=21993303&build=867956)

Issue #573, opened 2015-04-24
Add MultiscaleImage transformation 


prudhomm:
@LANTZT add support for MultiscaleImage  in Feel++ with the proper interface
in particular the coarse to fine relation : input is the coarse point coordinates and output is the value of the image at the fine level. This is need for #572.

prudhomm:
see 0b04c7f

Issue #572, opened 2015-04-24
Support new keyword : msi 


prudhomm:
@LANTZT we should support `msi` for multiscale image that generate a transformation between a coarse level to a fine level based on images.  `msi` is used in the dsel language to allow access to fine level image information and use at a coarse one.

PR #627, opened 2015-07-30
Holo3 Image Correlation

LANTZT:
This will update and introduce into Feelpp all the files develop during the internship linked to this subject.
It will also allow to update the version install on a server at Holo3, in order to be used later.

The resolution of the image correlation problem doesn't work right now and it's still "work in progress" .
#571 #572 #573 #586 #587


prudhomm:
how is it working now ?

LANTZT:
I manage to find the error which gave me difference values than the Octave/Matlab code.
Now I have the same result at each step, I test it on different easy functions ( x, sin(x), y*y, ...) 
But the visualization on Paraview didn't really change and it's still bad.

prudhomm:
Ok great ! what was it ? at which step ?
you see sometimes there are no ways to avoid this kind of hard search for the bugs

prudhomm:
now there might be another issue somewhere, you tested only the fft on each element no ?
or are you trying on the complete system ?
first thing is try simple functions, say `x` (gradient is easy)  and check using the `print` returns the proper values for the gradient

LANTZT:
Yep 
It was the construction of the waves number vector, the fftshift method I develop and use it didn't work well and return me something different than what I want and what is used in Octave.

LANTZT:
At this time, I used the FFT method on each element and I store all results in an array. Then I retrieve the value with real coordinates and reference one ( to define element for nodes at the boundary ).
The print didn't give me the good value in theory ( with  x for example ) but on Praview, it also different than what I can see .

Issue #653, opened 2015-09-10
Reduce mesh data structure memory footprint and improve loading


prudhomm:
We should improve loading time ofr a mesh data structure, to do so we can reduce the quantities stored by each element which would also reduce the memory footprint.
Instead of storing these quantities, we compute them on-demand without caching them for now

Issue #579, opened 2015-05-04
Support partitioning strategies 


prudhomm:
We should now support partitioning strategies in Feel++ rather than rely only on Gmsh
 - [x] metis
 
Other partitioner like parmetis and zoltan should probably be supported in the future

a specific library directory will be created to host the `Partitioner` classes

prudhomm:
@vincentchabannes we can close this issue no ?

PR #654, opened 2015-09-10
Feature/faster

prudhomm:
implements 
 - #653 reduce mesh memory footprint and improve loading time
 - #579 support partitioning strategies 
 - #657 improve timer support
 - #656 fix bilinear form assignement operator

 - [x] check `updateForUse` with respect to faces construction (still important time in HDF5 even though all faces are stores)
 - [x] make `updateForUse` work for skeleton meshes (submesh of faces,  edges or points)
 - [x] store only element ghost cells (no need for points,edges,faces ghosts, they are recomputed) 
 - [x] do not store barycenter faces, recompute on the fly
 - [ ] refactor `pidInPartition()` see geoentity
 - [x] deprecate `setNumberOfPartitions` in geoentity and do not store `numberOfPartitions()`

/cc @vincentchabannes @Trophime


prudhomm:
@vincentchabannes what do you think about 
 - removing by_ghostcell in elements, this can be done with id+pid and then for ghostcell. it is granted that it is nicer by_ghostcell but it means also more computational time for mesh generation and more storage. [1]
 - removing the by_pid index of faces, this is redundant with the composite one id+pid ? I did that for the elements already.


1. here is the list of files where GhostElement is used
```
find . -name "*.[hc]pp" | xargs -Istr grep -ln GhostElement str
./feeldiscr/meshimpl.hpp
./feeldiscr/operatorlagrangep1.hpp
./feelmesh/elements.hpp
./feelmesh/filters.hpp
./feelmesh/geoentity.hpp
./feelmodels/modelmesh/dofrelationshipmap.hpp
```

vincentchabannes:
@prudhomm ok for remove the by_pid in elements.
but  for the face, there is an important change : 
 - the access of face directly by the id will not be possible but only with couple id+pid.
agree with this?
In fact, this kind of access is already active for the element and it has been an additional difficulty in the parallelism implementation. Thus the change in faces must be  used with caution.

prudhomm:
Ok you will explain it to me tomorrow. For faces I agree this can be more delicate.

PR #660, opened 2015-09-16
implement the hdf5 format in save and load

prudhomm:
use type parameter : `u.save(_path=path, _type="hdf5")`


prudhomm:
@romainhild could you add a note in GitBook regarding saving/loading elements (Say in QuickReference/spaces.md somewhere)

romainhild:
@prudhomm would you want me to use markdown or asciidoc ?

prudhomm:
try asciidoc

Issue #527, opened 2015-02-11
Improve support for slepc 


prudhomm:
@romainhild we are going to improve the support for slepc and its option in feel++ and have something like we have for standard linear algebra problems

romainhild:
The following options should be supported : 
- [eps_mpd](http://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetDimensions.html)
- [eps_interval](http://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSSetInterval.html)
- [eps_krylovschur_partitions](http://slepc.upv.es/documentation/current/docs/manualpages/EPS/EPSKrylovSchurSetPartitions.html)

To implement `eps_interval`, we need to set the type to GHEP, the spectral transformation to SINVERT, the solver to KRYLOVSCHUR, the spectrum to EPS_ALL, get the [KSP](http://slepc.upv.es/documentation/current/docs/manualpages/ST/STGetKSP.html#STGetKSP) and set the ksp_type to preonly and the pc_type to cholesky.
We also need to add the possibility to use mumps for the preconditioner. 
An example is available [here](http://slepc.upv.es/documentation/current/src/eps/examples/tutorials/ex25.c.html).

romainhild:
we also need to add `-eps_target` which is used for the shift

prudhomm:
@romainhild should be closed no ?

romainhild:
Indeed, the main options are supported.
I guess we could support other types of eigen problems (nonlinear or polynomial are now supported by slepc) but they can have their own issue whenn needed.

PR #661, opened 2015-09-16
implement #527 : mpd and interval

prudhomm:
@prudhomm what do you think ?
still need to add krylovschurpartitions


prudhomm:
@romainhild could you start a section or a file in Gitbook regarding slepc and the associated interface ? features to solve eigenvalue problems ?
this can be used afterwards also in papers and reports

romainhild:
I'm trying with asciidoc.
For the eigen problem, can I add it to Solvers ? At the same level as Preconditioner and Backends ?

prudhomm:
yes definitely

Issue #479, opened 2014-12-02
CMake modification: FindFeel++.cmake


aancel:
@prudhomm We discussed this morning with @vincentchabannes and @jeromevelut about modyfing the way CMakefiles are built and how to use Feel++ as an external library.

Currently, the FindFeel++.cmake mainly checks for the dependencies of Feel++. We propose to rename the FindFeel++.cmake, e.g into feelpp.check.dependencies.cmake which would be included in the main CMakeLists.txt, and to create a more traditional FindFeel++.cmake or Feel++Config.cmake file, with the classical variables Feel++_FOUND, FEELPP_INCLUDE_DIR, etc.
This file would be configured during the CMake process (.in file).

vhuber:
@aancel On macOs I have compiled a custom version of Petsc
I have one installed with homebrew.
setting PETSC_DIR to the custom version does not prevent cmake to use the homebrew one.

aancel:
@vhuber Do you still have this issue ?

vhuber:
here the logs I can produce.
Please note I'm not confident in my petsc system installation with homebrew.
With no petsc_dir:
```sh
-- Petsc Dir 2: PETSC_DIR-NOTFOUND
-- Petsc Dir 1: /usr/local
-- PETSC_ARCH:
/tmp/titi/Makefile.petsc:4: /usr/local/lib/petsc/conf/rules: No such file or directory
/tmp/titi/Makefile.petsc:5: /usr/local/lib/petsc/conf/variables: No such file or directory
make: *** No rule to make target `/usr/local/lib/petsc/conf/variables'.  Stop.
/tmp/titi/Makefile.petsc:4: /usr/local/lib/petsc/conf/rules: No such file or directory
/tmp/titi/Makefile.petsc:5: /usr/local/lib/petsc/conf/variables: No such file or directory
make: *** No rule to make target `/usr/local/lib/petsc/conf/variables'.  Stop.
/tmp/titi/Makefile.petsc:4: /usr/local/lib/petsc/conf/rules: No such file or directory
/tmp/titi/Makefile.petsc:5: /usr/local/lib/petsc/conf/variables: No such file or directory
make: *** No rule to make target `/usr/local/lib/petsc/conf/variables'.  Stop.
/tmp/titi/Makefile.petsc:4: /usr/local/lib/petsc/conf/rules: No such file or directory
/tmp/titi/Makefile.petsc:5: /usr/local/lib/petsc/conf/variables: No such file or directory
[...]

PR #670, opened 2015-10-15
Feature/findfeelpp (Issue #479)

aancel:
Proposition for reworked FindFeel++.cmake:
- Move dependency checking into feelpp.dependencies.cmake
- Make FindFeel++.cmake a simple library finder 
- Add Feel++Config.cmake for project configuration (cf. [here](https://cmake.org/Wiki/CMake/Tutorials/How_to_create_a_ProjectConfig.cmake_file))
- External projects using the Feel++ library won't have to go through Feel++ dependency checking anymore as they are all bundled in the Feel++Config.cmake file


vhuber:
Suggestion: what about a `feelpp_options.cmake` that contains every `feelpp_*` options to include at the beginning of `feelpp.dependencies.cmake`. That will be easier for user to see the various options that are handled.

prudhomm:
@aancel ok seems an important contribution
 - everything recompiles as before ? 
 - how about sub-projects have you checked that it works too at least for two or three of them ?
 - how about checking on other platforms than atlas (eg. mesocenter, curie)?

aancel:
Yep, everything recompiles as before. The only change I made for the library is moving all the code from FindFeel++.cmake to feelpp.dependencies.cmake and including it back in the main CMakeLists.txt.
I'll be checking with subprojects and on other platform to ensure that it works as before.
Travis builds seem to be ok apparently.

prudhomm:
ok super, let me go how it goes with the sub-projects before I can merge

aancel:
Just a note for updating future projects:
The basic CMakeLists.txt to detect Feel++ is: 
```
if ( ${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR} )
  cmake_minimum_required(VERSION 2.8)
  FIND_PATH(FEELPP_CMAKE_MODULES FindFeel++.cmake PATH $ENV{FEELPP_DIR}/share/feel/cmake/modules /usr/share/feel/cmake/modules/ /usr/local/share/feel/cmake/modules/ )
  if ( FEELPP_CMAKE_MODULES )
    set(CMAKE_MODULE_PATH ${FEELPP_CMAKE_MODULES})
  else()
    message(FATAL_ERROR "Feel++ does not seem to have been installed on this platform")
  endif()
  OPTION(FEELPP_ENABLE_SYSTEM_EIGEN3 "enable system eigen3 support" ON)
  Find_Package(Feel++)
endif()
```
This will have to be replaced by:
```
find_package(Feel++ HINTS $ENV{FEELPP_DIR}/share/feel/cmake /usr/share/feel/cmake /usr/local/share/feel/cmake /opt/share/feel/cmake)
```

prudhomm:
@aancel could you document Gitbook with all this please ?

prudhomm:
Very nice

aancel:
@prudhomm I was just looking for info about how to build external project with Feel++, but I didn't find it. I'll add it again in the tutorial section

aancel:
@prudhomm What to you think about the idea of @vhuber about regrouping all the options of cmake ?

I also have another question: What do you think about enabling by default every project the user clones in the research directory ? It seems a rather logical idea, as if a user clones its project in research, it is because he/she wants to build it, and there is not much overhead apart from the cmake step.

I also changed some of the cmake files regarding to modules to ensure that modules are really correctly checked out.
Do you see other changes I should make ?

For me, It should soon be ok. I managed to build everything on atlas and curie (The mesocenter of Strasbourg would require a full module update to test it) and the modifications will only affect people building their projects outside of Feel++ source dir.

Issue #673, opened 2015-11-04
Support symmetric/SPD matrices and trigger associated solver/preconditioners 


prudhomm:
We should be able to handle SPD problems with solver/preconditioner for SPD problems
eg cg, cholesky, incomplete cholesky...
Note mumps supports the factorisation  L D L^T in parallel but there are some pitfalls to consider.

/cc @vincentchabannes @cdaversin @vhuber

PR #674, opened 2015-11-04
Feature/symm

prudhomm:
This PR implements:
 - #673 

@vincentchabannes could you review the code and tell me if that's OK with you ?


vincentchabannes:
@prudhomm yes it's ok for me.

prudhomm:
@aancel is there a problem with PETSc lookup ? 
see https://travis-ci.org/feelpp/feelpp/builds/89222432

it is bizarre because  it went through once. now it fails when looking for PETSc

aancel:
@prudhomm Yes, I made some modifications to be more informative and crash whenever needed, but it fails with travis. I'll revert that.

Issue #672, opened 2015-11-03
Exporter prefix


aancel:
Following the test written by @gdolle ([testsuite/feelts/test_bdf3.cpp](https://github.com/feelpp/feelpp/blob/develop/testsuite/feelts/test_bdf3.cpp)), we noticed some issues regarding the prefix used by the exporter:
An example would be (from @gdolle's code):
```
# First create an exporter with the test_bdf prefix
# This will create the initial timeset with the corresponding prefix
auto e = exporter( _mesh=mesh, _name="test_bdf" );
...
# Further calls to setPrefix won't modify the prefix of the timeset 
# after this initial call, because the setPrefix function does not update 
# the prefix of the latest timeset.
e->setPrefix( prefix )
```

Actually and currently, this is working with the Ensight Gold exporter because I made modifications to actually use the prefix set to the exporter and not the name of the timeset, but I think it would be a better idea to totally switch to using the name of the timeset. 

The function `setPrefix( name )` currently only store the value for the prefix. Wouldn't it be a better idea to directly set the value of the most recent timeset rather than storing the prefix ? Maybe we can also add a new `setPrefix( ts_id, name )` that would allow to modify the prefix for other timesets (in case we would be using several of them)

prudhomm:
I would have to have a closer look. I wrote that a long time ago. I need to refresh my memories.

aancel:
I can take care of this and the associated modifications in the exporters, if you validate the change. I think it would make the code of the exporters cleaner.

prudhomm:
sure, do that in a branch + PR

prudhomm:
@aancel Please discuss here or the branch/PR the changes before hand rather than after.

PR #675, opened 2015-11-04
Issue #672: Exporter prefix

aancel:
Proposition for issue #672 :
- [ ] Allow to change the name of a timeset instance (feel/feeldisc)
- [ ] Set the name of a timeset from the exporter with the setPrefix function (previously it was notably set when an instance of an exporter is created)
  - [ ] Maybe only allow to use the setName function from timeset to selected classes (`friend` ?)
  - [ ] Have a setPrefix( name ) function fo the last timeset, allow modification for other timsets with setPrefix( id, name )
- [ ] Update the exporters to only use the name of the timesets and not the prefix set in the bae exporter class


aancel:
@gdolle test_bdf3 is now working with this branch (both test1 and 2)

prudhomm:
@aancel could you post an example of the resulting case file generated if one timeset and multiple timeset ?

aancel:
@prudhomm It won't fundamentally change how the case file will be generated.
For example, when using @gdolle's test: test_bdf3, 2 case files will be output, one per timeset, whereas before the export was failing for the second timeset if I remember well, so you would have:
rhs_0.case
```
FORMAT:
type: ensight gold
GEOMETRY:
model: rhs_0.geo
VARIABLES:
scalar per node: 1 u0 rhs_0.u0.scl.****
TIME:
time set:        1
number of steps: 20
filename start number: 1
filename increment: 1
time values: 0.000000e+00 5.000000e-02 1.000000e-01 1.500000e-01 2.000000e-01 2.500000e-01 3.000000e-01 3.500000e-01 4.000000e-01 4.500000e-01
5.000000e-01 5.500000e-01 6.000000e-01 6.500000e-01 7.000000e-01 7.500000e-01 8.000000e-01 8.500000e-01 9.000000e-01 9.500000e-01
```
and 
```
FORMAT:
type: ensight gold
GEOMETRY:
model: rhs_1.geo
VARIABLES:
scalar per node: 1 u1 rhs_1.u1.scl.****
TIME:
time set:        1
number of steps: 20
filename start number: 1
filename increment: 1
time values: 0.000000e+00 5.000000e-02 1.000000e-01 1.500000e-01 2.000000e-01 2.500000e-01 3.000000e-01 3.500000e-01 4.000000e-01 4.500000e-01
5.000000e-01 5.500000e-01 6.000000e-01 6.500000e-01 7.000000e-01 7.500000e-01 8.000000e-01 8.500000e-01 9.000000e-01 9.500000e-01
```

It is more a matter of bugfixing rather than modifying how .case files are output.

aancel:
@gdolle Did you test the branch ?
@prudhomm What do you think of the changes ? Shall I continue to modify the exporters or do we drop the idea ?

prudhomm:
@aancel could you remerge with develop and let travis check that everything is ok with this PR ?

Issue #561, opened 2015-04-12
Support postprocessing in json files


prudhomm:
Json files now describe 
- material
- parameter
- boundary conditions

we should start supporting description of post processing such as : 
- fields : which fields to visualize 
- forces: on which boundaries should we compute stresses (eg force)
- points: evaluation of fields at points. List of points can be given either in 
  - json 
  - a filename.  the filename will be useful if there are many points
  - a marked entity in the mesh

``` json
 "Fields":["pressure", "velocity","vorticity"],
 "Forces": ["cylinder"],
 "Points":[ "pressure": [ [0.1,0.2], [0.15,0.2], .. ], }
  // or 
"Points":[ "pressure": { "filename": "pressure_at_points.dat"}
  // or 
"Points":[ "pressure": { "region": "cylinder"}
```

vhuber:
About materials:
- the present framework provides constant parameter, I would really enjoy to pass expression for them: the dynamic viscosity car be a function of a porous density.
- For the magnetostatic problems, we have to implement constitutives laws and tests various ones. It would be fantastic to use ginacExVf to define theses laws.

prudhomm:
I know.

vhuber:
I have added to the modelparameters class the ability to retrieve a node as a string in the tree.
That is a temporary workaround that allow to pass theses laws in the json file.

vhuber:
@vincentchabannes @prudhomm Is there at this point a way to pass a list of points (in a file) to evaluate ?
The current grammar will be a way too expensive when we have more than a very little set of points.

prudhomm:
this can easily be implemented and provided in the example of this issue

vhuber:
Ok, I'm working on it.

prudhomm:
I want a branch and a pull request, no direct commit please

vincentchabannes:
and please, check also that break nothing in feelmodels (see for example feelpp/applications/models/fluid/TurekHron/cfd3.json )

vhuber:
@vincentchabannes I mimic the framework you've implemented, but associating a field to a point sounds strange to me. what do you think of :
```javascript
        "Measures":
        {
            "firstSet":
            {
                "fields":["pressure","velocity"],
                "pointA":
                {
                    "coord":"{0.6,0.2,0}",
                },
                "pointB":
                {
                    "coord":"{0.15,0.2,0}",
                }
            }
            "secondSet":
            {
                "fields":["pressure"],
                "pointA":
                {
[...]

PR #687, opened 2016-01-08
[PostProcess] Provide a file for a list of points #561

vhuber:
I guess we still have to discuss about it. @prudhomm @vincentchabannes


prudhomm:
@vincentchabannes could you review this PR ?
thanks in advance

Issue #713, opened 2016-02-13
Support creating a VectorPetsc for a VectorUblas


prudhomm:
We should support creating a `VectorPetsc` out of a `VectorUblas` 
```c++
auto Xh = Pch<1>(mesh);
auto u=Xh->element();
auto v = toPETSc( u ); // get a VectorPetsc<>
auto up=Xh->elementPtr();
auto vp = toPETSc( up ); // get a shared_ptr<VectorPetsc<>>
auto vp = toPETScPtr( u ); // get a shared_ptr<VectorPetsc<>>
```
also related to #393 
/cc @vincentchabannes @prj-

PR #714, opened 2016-02-13
Implements #713 Ublas to Petsc interface

prudhomm:
implement creating a VectorPetsc out of a VectorUblas without copying
@vincentchabannes @prj-


feel/feelalg/vectorpetsc.hpp
```diff
@@ -170,6 +171,23 @@ class VectorPetsc : public Vector<T>
...
+    template<typename Storage>
+    VectorPetsc( VectorUblas<T,Storage>& v )
+        :
+        super( v.mapPtr() ),
+        M_destroy_vec_on_exit( false )
+        {
+            VecCreateMPIWithArray( v.map().worldComm().globalComm(),
+                                   1, v.map().nLocalDofWithoutGhost(), v.map().nDof(),
+                                   std::addressof( v[0] ), &M_vec );
+            this->M_is_initialized = true;
```
vincentchabannes: @prudhomm I think it's not good in parallel. the size of M_vec is nLocalDofWithoutGhost() but v is nLocalDofWithGhost(). And the ordering is not the same.

prudhomm: what do you mean by ordering ?

prudhomm: actually there is another function that would include the ghosts
http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecCreateGhostWithArray.html#VecCreateGhostWithArray

prudhomm: so `[0,nLocalDofWithoutGhost()] ` does not correspond to the indices for the local array without the ghosts ?

vincentchabannes: yes, it is what I mean.
There is an additional Vec M_vecLocal with the same size and ordering. It's used for the container access with local dof index (and allow copy from VectorPetsc to VectorUblas). This vector is updated from M_vec in Localize() method with MpiScatter. This strategy is old (at the begin of feel++ //) and can be rethought.

vincentchabannes: Maybe we can remove  M_vecLocal and build M_vec with VecCreateGhostWithArray

prudhomm: but then how about VecCreateGhostWithArray ?

prudhomm: and actually you can access the ghost part of a `vec` using  `VecGhostGetLocalForm() `

vincentchabannes: I looked quickly the doc and example of VecCreateGhostWithArray. Indeed, it seems to be the good way (and better that is done actually)

prudhomm:
@vincentchabannes could you have a look at the implementation ? there is currently a crash
do we have access to ghosts global ids ? do we actually need them ?
from petsc's doc we can provide the global indices of ghost points (or NULL if not needed), these do not need to be in increasing order (sorted)

prudhomm:
I am currently testing on the laplacian in quickstart

PR #725, opened 2016-03-08
Feature/cnab2

jbwahl:
@prudhomm


prudhomm:
so it works too with aerothermal flows ?

prudhomm:
how about 3D ?

jbwahl:
it works with aerothermal but no validation for now
3D coming soon...

prudhomm:
done

PR #745, opened 2016-04-13
Feature/crb clean

prudhomm:
cleanup CRB before refactoring


prudhomm:
@vincentchabannes I created a PR

prudhomm:
@cdaversin could you check that your codes are still running properly in this branch. also could you tell us why there are finite element matrices built also in the online case ? is it for the offline part of SER ?

cdaversin:
@prudhomm I will try to run my codes in this branch as soon as possible and keep you informed. Concerning the finite element matrices you speak about, can you give more precisions ? During online step, the RB output can be compared with the parametric FE one which supposes the resolution of a FE dimension problem. The computation of the FE output is handled by the option `crb.compute-fem-during-online` (default value is true).

prudhomm:
it is in the feelpp/crb gitter salon 
here it is also https://github.com/feelpp/feelpp/blob/develop/feel/feelcrb/crb.hpp#L9148

prudhomm:
Merged

Issue #759, opened 2016-05-04
CRB / PC


vhuber:
The preconditioners (primal, dual, l2) are members of the CRB class.
That forbids the user to define its own PC for theses methods.

We should - if there is no need to keep them - remove theses members of the class.

@feelpp/crb

prudhomm:
following the various discussions on gitter I close the issue

PR #760, opened 2016-05-04
Remove PC from crb

vhuber:
#759


prudhomm:
@vhuber use #766 instead and `attachInHousePreconditioner()`

PR #761, opened 2016-05-04
add postPro virtual function to crbmodel

vhuber:
In the magnetostatic case, the system is solved for the potential vector, not the magnetic field.
I need to apply the `Curl` operator to export the correct value.

The postPro function allow the user to manipulate it's results after a fem/rb evaluation.


prudhomm:
that's an interest usecase. I will look at the magnetostatic code. 
it should be renamed `postProcess`

prudhomm:
@vincentchabannes @cdaversin could you review this PR? 
I just merged with `develop` and fixed a conflict with in `OpusApp`
thanks in advance !

Issue #776, opened 2016-05-13
Support Altair development environment


prudhomm:
/cc @feelpp/po

prudhomm:
@vincentchabannes the `feelpp_database_*` have been installed in `feelpp-libs`. Could you test if it works properly ? could you add to feel++ the acusim data @StephieEDWIGE produced to test the tools ?

prudhomm:
@feelpp/po the conversion, pod and export tools are now distributed via docker

vincentchabannes:
@prudhomm I have this error in the docker container for all applications : 
```
feelpp_qs_laplacian_2d: error while loading shared libraries: libh3dreader.so: cannot open shared object file: No such file or directory
```

prudhomm:
I fixed that in #895, I will rebuild as soon as I get the automatic dockerization ready (I am close now I think)

PR #777, opened 2016-05-13
Feature/altair

prudhomm:
implements 
- #776


prudhomm:
@vincentchabannes it does not compile on buildkite, see error message here
https://buildkite.com/feelpp/feelpp/builds/1305#_

PR #795, opened 2016-05-30
Feature/glog crbjson

prudhomm:
@vincentchabannes @cdaversin I have merge feature/glog and feature/crb-json
@vincentchabannes could you verify that everything is all right for you ?
have you actually change `OpusApp::run()` ? it was moved in `opusapp_impl.hpp`


prudhomm:
@vincentchabannes any feedback on the logs ?

I was thinking about adding
 - appdir default to `appName()/np_<xy>`  
 - logdir default to $appdir/logs

both are options that can be changed in the command line.
moreover 
 - `--no_log` get rids of the logs 
 - `--npdir=false` removes `np_<xy>`

what do you think ?

vincentchabannes:
@prudhomm about the log, my paraview plugin make log directory everywhere on my disk (place where I run paraview).  the --no_log option doesn't seem to work.
I need to say at Environment build that I don't want to write the log file or specify the directory where log are writed (not necessary from command line option).
For feelmodels app, I need to think more about that.

For my point of view, I would prefer to disable by default the logfiles (only activated with a command line option). Also, this can avoid some io disturb in supercomputer ( depending of traffic, this writing can have a cost).

PR #824, opened 2016-12-02
Add support for MESH and MED mesh format

Trophime:
The idea is to get support for mesh format supported by gmsh.
The code is based on example from gmsh api in utils/api_demos/mainReadMesh.cpp
and Common/OpenFile.cpp.

A test in testsuite allow to check this new fonctionnality for MED.
You may need to change the default MED mesh filename.

### DIFF
--- a/feel/feelfilters/loadMEDmesh.hpp
+++ b/feel/feelfilters/loadMEDmesh.hpp
@@ -22,12 +22,12 @@
   Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 */
 /**
-   \file loadgmshmesh.hpp
-   \author Christophe Prud'homme <christophe.prudhomme@feelpp.org>
-   \date 2013-12-24
+   \file loadmeshmesh.hpp
+   \author Christophe Trophime <christophe.trophime@lncmi.cnrs.fr>
+   \date 2016-11-30
  */
-#if !defined(FEELPP_LOADGMSHMESH_HPP)
-#define FEELPP_LOADGMSHMESH_HPP 1
+#if !defined(FEELPP_LOADMEDMESH_HPP)
+#define FEELPP_LOADMEDMESH_HPP 1
 
 #include <feel/feelfilters/gmsh.hpp>
 #include <feel/feelfilters/detail/mesh.hpp>
@@ -46,7 +46,7 @@ namespace Feel {
  */
 BOOST_PARAMETER_FUNCTION(
     ( typename Feel::detail::mesh<Args>::ptrtype ), // return type
-    loadGMSHMesh,    // 2. function name
+    loadMEDMesh,    // 2. function name
 
     tag,           // 3. namespace of tag types
 
@@ -73,6 +73,8 @@ BOOST_PARAMETER_FUNCTION(
       )
                          )
 {
+    cout << "loadMEDmesh:\n";
+
     typedef typename Feel::detail::mesh<Args>::type _mesh_type;
     typedef typename Feel::detail::mesh<Args>::ptrtype _mesh_ptrtype;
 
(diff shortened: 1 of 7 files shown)


Trophime:
One comment: Of course it requires that gmsh hasbeen built with med support
This should be properly dected within feel/cmake/modulesFindGmsh.cmake.
This is the case for Debian/Ubuntu.

However there is still an issue with Salome. Recent Salome version ships with 3.20 MED library (aka med-fichier source package in Debian/Ubuntu). This version is not available at the moment in Debian/Ubuntu.

Trophime:
As for the failed build It works on my Debian testing box with clang 3.9.
I don't understand why it fails on Ubuntu 16.10 with clang 3.9 (actually it's feelpp library that FTBS).

prudhomm:
probably too much memory is required :(
I will look into this

feel/feelfilters/loadMEDmesh.hpp
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * \arg mesh mesh data structure
+ * \arg filename filename string (with extension)
+ * \arg refine optionally refine with \p refine levels the mesh (default: 0)
+ * \arg update update the mesh data structure (build internal faces and edges) (default : true)
+ * \arg physical_are_elementary_regions boolean to load specific meshes formats (default : false)
+ */
+BOOST_PARAMETER_FUNCTION(
+    ( typename Feel::detail::mesh<Args>::ptrtype ), // return type
+    loadMEDMesh,    // 2. function name
```
prudhomm: why not merge with loadMesh ?

feel/feelfilters/loadMESHmesh.hpp
```diff
@@ -0,0 +1,132 @@
...
+ * \brief load a mesh data structure (hold in a shared_ptr<>) using GMSH
+ *
+ * \arg mesh mesh data structure
+ * \arg filename filename string (with extension)
+ * \arg refine optionally refine with \p refine levels the mesh (default: 0)
+ * \arg update update the mesh data structure (build internal faces and edges) (default : true)
+ * \arg physical_are_elementary_regions boolean to load specific meshes formats (default : false)
+ */
+BOOST_PARAMETER_FUNCTION(
+    ( typename Feel::detail::mesh<Args>::ptrtype ), // return type
```
prudhomm: idem here, there is redundant code that may be difficult to maintain in the end

testsuite/feelfilters/test_med.cpp
```diff
@@ -0,0 +1,121 @@
...
+        Feel::cout << "Number of faces in mesh : " << mesh->numFaces() << "\n";
+        Feel::cout << "Number of edges in mesh : " << mesh->numEdges() << "\n";
+        Feel::cout << "Number of elts in mesh : " << mesh->numElements() << std::endl;
+
+        Feel::cout << "Markers:" << std::endl;
+        for (auto marker:  mesh->markerNames() )
+        {
+            auto name = marker.first;
+            auto data = marker.second;
+            Feel::cout << "\t" << name << " dim=" << data[1] << " (" << name.size() << ")\n";
```
prudhomm: why not add checks that the marker have been properly read ?

prudhomm:
@Trophime I think we need to redesign a tad the `loadMesh` function, it won't scale (boost.parameter code come with a cost, we should be careful) with new formats. I have some ideas.

feel/feelfilters/loadMEDmesh.hpp
Trophime: I agree that would be great. The only point is that MED is a specific case...
for other extension there is a correspondance between the extension and a readXXX method in Gmsh.

But from a practical point of view I've no idea how to code that (beside having some kind dictionnaries linking extension and pointer to the proper readXXX method )

Issue #828, opened 2016-12-28
update Eigen3 


prudhomm:
Eigen 3 should be upgraded to latest developments from bitbucket

prudhomm:
here is the result of running fsi_2d 
````
finish-step1 in 4.00003e-06
finish-step2 in 1.00024e-06
finish-step3 in 0.000113
 it 2  nbLocalisationFail 0

||==============================================||
||==============================================||
||==============================================||
||----------Info : FluidMechanics---------------||
||==============================================||
||==============================================||
||==============================================||
   Prefix : fluid
   Root Repository : /home/prudhomm/feel/applications/models/fsi/TurekHron/fsi3/P2P1G1-P1G1/np_4
   Physical Model
     -- pde name  : Navier-Stokes
     -- stress tensor law  : newtonian
     -- time mode : Transient
     -- ale mode  : Yes
     -- gravity  : false
   Physical Parameters
     -- rho : 1050
     -- mu  : 0.00345
     -- nu  : 3.28571e-06
   Boundary conditions
       -- Dirichlet (elimination) : fluid-cylinder , fluid-inlet , fluid-wall
       -- fluid outlets [free] : fluid-outlet
       -- MeshALE[fixed] : fluid-cylinder , fluid-inlet , fluid-wall , fluid-outlet
       -- MeshALE[moving] : fsi-wall
   Space Discretization
     -- mesh file name   : /home/prudhomm/feel/applications/models/fsi/TurekHron/fsi3/P2P1G1-P1G1/meshes/fsi_fluid_Simplex_2_1_2_M1_p4.json
     -- nb elt in mesh  : 2654
     -- geometry order  : 1
     -- velocity order  : 2
     -- pressure order  : 1
     -- nb dof (u+p)    : 12498 (3179)
     -- nb dof (u)      : 11060
     -- nb dof (p)      : 1438
     -- stabilisation   : OFF
   Time Discretization
     -- initial time : 0
     -- final time   : 10
     -- time step    : 0.005
     -- order        : 2
     -- restart mode : No
     -- save on disk : true
     -- freq save : 1
     -- file format save : binary
   Exporter
     -- type            : ensightgold
     -- high order visu : OFF
     -- freq save       : 1
     -- fields exported : velocity - pressure - displacement - pid
   Processors
     -- number of proc environnement : 4
     -- environement rank : 0
     -- global rank : 0
     -- local rank : 0
   Numerical Solver
     -- solver : Oseen
||==============================================||
[...]

PR #829, opened 2016-12-28
Feature/eigen3

prudhomm:
implments #828


prudhomm:
see #828 for more details

Issue #839, opened 2017-01-18
Reorganize models into toolboxes


prudhomm:
In order to develop #834 we need to reorganize the models by introducing a toolboxes directory.
The models libraries and applications move there.
Some features
 * enables #834 and hence facilitate docker implementation
 * enable/disable toolboxes from cmake
 * one place for the toolboxes

Issue #834, opened 2017-01-07
Support staged compilation and installation 


prudhomm:
Provide support for different levels/stages
 . libs 
 . Base
 . Apps
 . Crb apps
Currently we build in a monolithic manner

PR #840, opened 2017-01-18
Feature/toolboxes

prudhomm:
implements
 - #839 

enables #834


prudhomm:
@vincentchabannes could you have a look ?

PR #846, opened 2017-01-30
Fix missing target for project test

gdolle:
Add missing targets for feelpp_add_test function


prudhomm:
This has been fixed in another commit. It was not the right fix but thanks anyway for doing this !

PR #847, opened 2017-01-31
feature/meshStructured

LANTZT:
- Add another MeshStructured constructor that use nodes coordinates data from hbf files
- Used in Holo3 project for surface_rec application


LANTZT:
@prudhomm Yes indeed
I think I see how to pass trough some issues we can have by assembling the two constructors

PR #865, opened 2017-03-10
Feature/MeshStructured

LANTZT:
- Add possibility to build "deformed" mesh from coordinates files
- Add polygon selection of "bad" areas we don't want to take in account

### DIFF
--- a/feel/feeldiscr/meshstructured.cpp
+++ b/feel/feeldiscr/meshstructured.cpp
@@ -22,11 +22,17 @@
  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 #include <feel/feeldiscr/meshstructured.hpp>
+#include <boost/geometry.hpp>
+#include <boost/geometry/geometries/point_xy.hpp>
+#include <boost/geometry/geometries/polygon.hpp>
 
+typedef boost::geometry::model::d2::point_xy<double> poly_point_type;
+typedef boost::geometry::model::segment<poly_point_type> segment_type;
+typedef boost::geometry::model::polygon<poly_point_type> polygon_type;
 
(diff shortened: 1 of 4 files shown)


feel/feeldiscr/meshstructured.cpp
```diff
@@ -22,11 +22,17 @@
  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 #include <feel/feeldiscr/meshstructured.hpp>
+#include <boost/geometry.hpp>
```
prudhomm: Is there a requirement on Boost and Boost.geometry version ?

prudhomm: We should probably add a note on the use of Boost Geometry in Feel++ documentation. this is new definitely

feel/feeldiscr/meshstructured.cpp
```diff
@@ -72,6 +78,40 @@ MeshStructured::MeshStructured( int nx, int ny, double pixelsize, holo3_image<fl
...
     node_type coords( 2 );
 
+    
+    // polygon build
+    std::vector<polygon_type> list_poly ;
+    //std::cout << list_poly.size() << std::endl;
+    polygon_type poly;
+    bool inPoly = false;
+    bool testInPoly = false;
+    if (withPoly)
```
prudhomm: @LANTZT is this (polygonal extraction) described in the documentation of the holo3 project ?

prudhomm: Moreover is polygonal mesh extraction tested in the testsuite ?

LANTZT: Yes at the first question 
And No to the second

PR #869, opened 2017-03-16
add new version for non-linear benchmark (grepl), with parametrized rhs

cdaversin:
The only difference between `feature/glog-crbjson` branch and `develop` is my commit adding a new version of Grepl benchmark with a "parametrizable" rhs. This change could be added to develop, and we the `feature/glog-crbjson` branch could be closed. @prudhomm @vincentchabannes What do you think ?

### DIFF
--- a/applications/crb/BenchmarkGrepl/CMakeLists.txt
+++ b/applications/crb/BenchmarkGrepl/CMakeLists.txt
@@ -35,10 +35,16 @@ foreach(O 1 2 3)
 
   foreach(D 2 3)
     configure_file( benchmarkgrepl-nonlinear-elliptic.xml.in ${CMAKE_CURRENT_SOURCE_DIR}/benchmarkgrepl-nonlinear-elliptic${O}-${D}D.xml.in COPYONLY )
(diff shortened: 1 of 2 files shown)


prudhomm:
@vincentchabannes @romainhild @jbwahl we have to integrate @cdaversin work ! I completely forgot about that.
@cdaversin we have a hard time reproducing your results. Is that possible that some `.cfg` are not committed ?

cdaversin:
@prudhomm From what I remember, the ```cfg``` generated by the commited ```cfg.in``` was sufficient (I may had made some minor changes in the generated config file, but unfortunately I cannot check... )

prudhomm:
@cdaversin thanks !

jbwahl:
@cdaversin in `cfg`file you use `error_type=2` which is `NO_RESIDUAL`. So you used randomly chosen parameters for the construction of the reduced basis ? No error indicator ? 
(sorry to bother you with that)

cdaversin:
@jbwahl This ```cfg``` is auto-generated by the ```cfg.in```, with ```error_type=2``` by default. But I tried in different configurations, especially with various SER variants using error indicators. You may find  some results in the Notes I wrote as I was testing the variants. Also, there are some scripts using the corresponding options in the hifimagnet repo (as an example).

PR #871, opened 2017-03-21
Feature/minor fixes from imft

Doyeux:
I just re-installed a new environment for Feel and I ran on the same issue that I used to few weeks ago, being:

with a Ubuntu16.04 install, I got a segfault in tic() function.
It is due to a conflict with the static variable `local` defined to store timers. Changing the name `local` fixed the problem.

Moreover, few `close()` were missing after Jacobian construction in doc/manual/nonlinear and also led to sagfault.

### DIFF
--- a/doc/manual/nonlinear/bratu.cpp
+++ b/doc/manual/nonlinear/bratu.cpp
@@ -77,6 +77,7 @@ main( int argc, char** argv )
                             ( - trans( id( v ) )*( gradt( u )*N() )
                               - trans( idt( u ) )*( grad( v )*N() )
                               + penalbc*trans( idt( u ) )*id( v )/hFace() ) );
+            J->close();
         };
     auto Residual = [=](const vector_ptrtype& X, vector_ptrtype& R)
         {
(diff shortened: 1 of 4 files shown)


prudhomm:
@Doyeux thanks. merged !

vincentchabannes:
The close() added are strange. Normally we don't need to call close here, it's automatically done (see https://github.com/feelpp/feelpp/blob/develop/feel/feelalg/solvernonlinearpetsc.cpp#L246), so it's call two times in the app. I will try to reproduce and fix the issue.

Doyeux:
Ok, tell me if you can reproduce.
I had the bug on Ubuntu16.04 with two different versions of petsc.

Doyeux:
The fix 
https://github.com/feelpp/feelpp/pull/871/commits/b35094cdf5975b355e83d1dc08ab68a0214a610b
didn't really fixed the problem. It seems the issue is more complex than a simple conflict of variable names.
I'm not sure how to exactly reproduce the issue. It happens with gcc 5 and Ubuntu16.
Also, going through the conversations in gitter, I saw someone already experienced it:
https://gitter.im/feelpp/feelpp?at=588b27cd1e4d4bd9629cbc0b

PR #888, opened 2017-04-30
Feature/cmake dependencies

vincentchabannes:
- up/clean detection of feel++ dependencies
- fix/refactoring of install process


cmake/modules/feelpp.module.nlopt.cmake
```diff
@@ -21,7 +21,7 @@
 # Foundation, Inc., 675 Mass Ave, Cambridge, MA 02139, USA.
 #
 #
-OPTION( FEELPP_ENABLE_NLOPT "Enable NLOPT (NonLinear Optimisation Library)" ON )
+option( FEELPP_ENABLE_NLOPT "Enable NLOPT (NonLinear Optimisation Library)" OFF )
```
prudhomm: why is it off ?

prudhomm: nlopt is shipped with Feel++. it should be off only in the minimal case IMO

vincentchabannes:
@prudhomm It's not finish and not fully tested.

PR #894, opened 2017-05-06
Feature/cmake dependencies

vincentchabannes:
- reorgonize hearder/config of feel++ lib install : use feelpp subdir. Warning, it change the path of the feel++ detection. It can be done for example : 
```
   find_package(Feel++ PATHS $ENV{FEELPP_DIR}/share/feelpp/feel/cmake/modules )
```
- up Asciidoctor detection : check backends and re-enable by default if detected 
- install tooboxes headers in <FEELPP_DIR>/include/feelpp/tooboxes
 - testcases install use also feelpp subdir (@prudhomm we need to update in docker the symbolic link of Testcases)
 - refactoring of cling interpreter script generation : this script in now generated with the target ```feelpp_cling_interpreter```


prudhomm:
@vincentchabannes I have modified the docker build accordingly to the install changes. it might still be broken, I will try to fix things today

vincentchabannes:
@prudhomm thanks!

prudhomm:
Some minor details need fixing. Almost done. BTW Altair comes with standard docker. I removed most of acusolve and keep only what we need. I will be able to finish the docker for felpp/po

Issue #1041, opened 2017-09-07
Bug with expansion


romainhild:
in `Feel::expansion`, we check that the size of the vector is less than the number of elements to use. It should of course be the other way, if `c.size() < M`, the line 52 will produce a segfault when reaching `i=c.size()`.
https://github.com/feelpp/feelpp/blob/80d98836fa907fd965fb829a3b37875bfc1bf8cb/feel/feeldiscr/expansion.hpp#L46-L53

Also in crb.hpp, we check that the N is greater than 0, or use the maximum number of basis, but after we use the N given in parameters:
https://github.com/feelpp/feelpp/blob/80d98836fa907fd965fb829a3b37875bfc1bf8cb/feel/feelcrb/crb.hpp#L9235-L9244
And is it normal at line 9242 that we have `Wn.size()` ? Shouldn't it be `u.size()` ?

romainhild:
@prudhomm @vincentchabannes no objections about this ?

vincentchabannes:
@romainhild ok for me

PR #1042, opened 2017-09-07
fix issue #1041

romainhild:
- Fix expansion check
- use min size of coefficients and elements


jbwahl:
It's ok for me, but i'm pretty sure that the `CHECK` in `expansion()` is useless since you change `M` before to be smaller than `c.size()` and `b.size()`.

romainhild:
indeed we can remove the check.

PR #1043, opened 2017-09-13
Fix py3k PYTHON_VERSION detection

jschueller:
In Python3 print becomes a function and cmake throws.

-----


prudhomm:
@jschueller thank you for the contribution.
are you interested in using the python interface for reduced basis ?

jschueller:
Just curious, was just trying to build that beast. Does it support Python3 ?

prudhomm:
I haven't tried. I need to check. 
We are preparing some demo for the ANR Chorus Workshop.
I will send you a link to Docker images with the python code running.

Note that in the docker images, I remember that both py2 and py3 are available (need to be extra careful with DLL..)

PR #1050, opened 2017-09-20
Feature/crb plugin

vincentchabannes:
- Update CRB plugin in order to support several model instantiations in the same library. The model name appers in the dll_alias :
 create_plugin -> create_plugin_"modelname"

- modification on the mast argument of  FEELPP_CRB_PLUGIN macro :  model name  must not be given between quote

- benchmarkgrep linear and nonlinear are now supported by the crb plugin framework

- up mongocxx commit version : some changes about libBson and libMongoC detection


vincentchabannes:
SER should work. I will test.
But there are some problems with the uid and also the modelname for each level

prudhomm:
we should add an issue and discuss the solutions. What do you think ? I remember that we discussed that some time ago but I don't quite remember the details

vincentchabannes:
Yes I will try to do the point about SER (asap) and I will fill an issue.

PR #1053, opened 2017-09-21
Feature/issue1052

romainhild:
- change `M_physics` to `std::set`
- add methods:
  - `physics()`: return the set of physics
  - `physic()`: return the first physic or "" if void
  - `setPhysics(std::set<std::string>)`: set the set of physics
  - `addPhysics(std::string)`: add one physic
  - `hasPhysics()`: return if the material has one or more physics
  - `hasPhysics(std::string p)`: return if the material has the physic `p`
  - `materialWithPhysics(std::string p)`: return a sub map of materials with physic `p`
- add const specifiers for getter methods
- change toolboxes to use first physic


vincentchabannes:
@romainhild thanks

Issue #1064, opened 2017-10-03
Merge hifimagnet toolboxes


romainhild:
We need to have all the toolboxes in the same place to be able to:
- merge the non-linear part of the thermoelectric toolbox of hifimagnet into the one of feelpp
- update the magnetostatic toolbox of hifimagnet to match the other toolboxes of feelpp

prudhomm:
Indeed we close the issue.

Trophime:
:+1:

PR #1065, opened 2017-10-03
fix issue #1064

romainhild:
We need to have all the toolboxes in the same place to be able to:

* merge the non-linear part of the thermoelectric toolbox of hifimagnet into the one of feelpp
* update the magnetostatic toolbox of hifimagnet to match the other toolboxes of feelpp


Trophime:
@prudhomm it seems I no longer have access to https://buildkite.com/feelpp/feelpp...

romainhild:
@prudhomm @vincentchabannes @gdolle this branch is ready to merge.
Do we still want it ?

prudhomm:
@romainhild  I think we still want it, we want to integrate the magnetostatic code as well as the other ones and  merge with the feelpp toolboxes.

@vincentchabannes ?

prudhomm:
@romainhild  @Trophime  @vincentchabannes  I merged develop in merge/hifimagnet, let's see what happens.

@romainhild is there a documentation regarding the magnetostatic toolbox?

romainhild:
Indeed we don't need the thermoelectric part anymore, but this pull request was made before.
We have some test indeed in hifimagnet.

For the magneto, the idea was not to keep it that way, but to have it next to the other toolboxes in order to be able to change it to something similar to the others.
The main problem is that it uses the MagnetTools from hifimagnet, for example to set a Dirichlet condition. We need to be able to compile it with or without the MagnetTools.

prudhomm:
At least we have something with the magneto toolbox, we can start from there. This needs to work for hifimagnet and mso4sc. 
@Trophime  @romainhild can we now use the thermo-electric toolbox for hifimagnet, have you tested it? my understanding is that, it should now support all features required for hifimagnet (multi-materials + non-linear).
If that works, then we only extract the magneto part and have it as a toolbox for now until we have a proper implementation. 
That way we have all the toolboxes in feel++  and hifimagnet puts them together to solve the hifimagnet problems

prudhomm:
@romainhild @Trophime  I would suggest that the magnetostatic part is extracted 

Moreover, have you tested the thermo-electric toolbox model? have you tried that on an LNCMI testcase?
I would think that we need to do that very soon with MSO4SC ending

PR #1089, opened 2017-11-08
Fix1063

Trophime:

-----


Trophime:
I've just re-enabled the check for some specific headers (OpenFile.hpp,..) that are needed
for the meshadaptation example in doc/manual/adapt. We can of course only use FEELPP_HAD_GMSH_ADAPT to simplify but it will be nice to keep a message to notify the user.

As for the second part of the patch, I guess you mean the changes in feel/feelfilters/importergmsh.cpp, there are needed in order to properly load med files!! Without these changes the mesh read from med is "empty".

prudhomm:
I merged the branch into develop

Issue #1051, opened 2017-09-20
Add ModelOutputs class to give the outputs of a CRB model in json file


romainhild:
If we want to have CRB models as independant as possible of the geometry, we should give the outputs to compute in a json file, each type of output would have a list of marker on which to compute it.
The syntax would be something like:
```json
"Outputs":
{
  "intensity":
  {
    "marker1":2,
    "marker2":3
  },
  "avg_temp":
  {
    "omega":3
  }
}
```
where the number is the dimension of the marker.

If anyone can think of something else an output can depend on, please tell me.

prudhomm:
`intensity` of what ? you have to pass a field 
I think you should have a subsection on the marker 
```
  "marker1":
   {
      "topodim": 3,
      "field(s)": [""] // field or fields to be used by the output 
      "... // may be more in the future depending on the output
   } 
```

I think that outputs could be defined in a "generic" way for some and "specific" way depending on the toolbox being used. 
For example `mean` is generic. `flux` too. we _just_ need to know the field and the marker. but toolboxes may also have specific output such as `tresca` or `vonmises` for elasticity

@romainhild @vincentchabannes what do you think ?

a remark: Ensight has a CAlculator. It would be interesting to look at their outputs and their configuration. It might give us some ideas

prudhomm:
An Output maps to a mathematical object which is then available through the toolboxes. 
We have generic scalar outputs (eg `min` `max` `mean` `flux` ...) that can be shared between toolboxes and specific outputs to one or several toolboxes

prudhomm:
My comments apply to both fem and crb

romainhild:
A possible interface for the json would be:
```json
Outputs:
{
  "myoutput":
  {
    "type":"min_value",
    "field":"potential",
    "range":"omega",
    "topodim":3
  },
  "output1":
  {
    "expr":"-sigma*grad(potential)*trans(grad(potential))",
    "field":"potential",
    "range":["surface1","surface2"],
    "topodim":2
  }
}
```
And we would have a base class `ModelOutput` which would do the computation in the case where we give it an expression. And several subclasses, for each type of output (average value, min value, max value, flux,...)
```c++
class ModelOutput
{
  std::string M_type, M_expr;
  std::vector<std::string> M_markers, M_fields;
  int M_topodim;

  template<typename T>
  virtual vector_ptrtype apply( T functionSpace)
  {
    auto range = markedfaces(functionSpace->mesh(), M_markers);
[...]

PR #1099, opened 2017-11-28
add modeloutput class to manage crb output

romainhild:
- add ModelOutputs to ModelProperties
- ModelOutputs is just a map of <string,ModelOutput>
- ModelOutput is just a class containing a name, a type, a range and a dimension
- change testsuite of modelproperties to check basic data

First draft for #1051

### DIFF
--- a/feel/feelcore/environment.cpp
+++ b/feel/feelcore/environment.cpp
@@ -2211,7 +2211,6 @@ Environment::expand( std::string const& expr )
     boost::replace_all( res, "${datadir}", dataDir );
     boost::replace_all( res, "${exprdbdir}", exprdbDir );
     boost::replace_all( res, "${h}", std::to_string(doption("gmsh.hsize") ) );
-    boost::replace_all( res, "${testsuite_srcdir}", topSrcDir + "/testsuite/" );
 
     boost::replace_all( res, "$feelpp_srcdir", topSrcDir );
     boost::replace_all( res, "$feelpp_builddir", topBuildDir );
@@ -2225,7 +2224,6 @@ Environment::expand( std::string const& expr )
     boost::replace_all( res, "$datadir", dataDir );
     boost::replace_all( res, "$exprdbdir", exprdbDir );
     boost::replace_all( res, "$h", std::to_string(doption("gmsh.hsize") ) );
-    boost::replace_all( res, "$testsuite_srcdir", topSrcDir + "/testsuite/" );
 
 
     typedef std::vector< std::string > split_vector_type;
--- a/feel/feelcore/environment.cpp
+++ b/feel/feelcore/environment.cpp
@@ -2211,6 +2211,7 @@ Environment::expand( std::string const& expr )
     boost::replace_all( res, "${datadir}", dataDir );
     boost::replace_all( res, "${exprdbdir}", exprdbDir );
     boost::replace_all( res, "${h}", std::to_string(doption("gmsh.hsize") ) );
+    boost::replace_all( res, "${testsuite_srcdir}", topSrcDir + "/testsuite/" );
 
     boost::replace_all( res, "$feelpp_srcdir", topSrcDir );
     boost::replace_all( res, "$feelpp_builddir", topBuildDir );
@@ -2224,6 +2225,7 @@ Environment::expand( std::string const& expr )
     boost::replace_all( res, "$datadir", dataDir );
     boost::replace_all( res, "$exprdbdir", exprdbDir );
     boost::replace_all( res, "$h", std::to_string(doption("gmsh.hsize") ) );
+    boost::replace_all( res, "$testsuite_srcdir", topSrcDir + "/testsuite/" );
 
 
     typedef std::vector< std::string > split_vector_type;
--- a/feel/feelmodels/modeloutputs.cpp
+++ b/feel/feelmodels/modeloutputs.cpp
@@ -1,0 +1,134 @@
+/* -*- mode: c++; coding: utf-8; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4; show-trailing-whitespace: t  -*-
+ 
+ This file is part of the Feel++ library
+ 
+ Author(s): Christophe Prud'homme <christophe.prudhomme@feelpp.org>
+ Date: 16 Mar 2015
+ 
+ Copyright (C) 2015 Feel++ Consortium
+ 
+ This library is free software; you can redistribute it and/or
+ modify it under the terms of the GNU Lesser General Public
+ License as published by the Free Software Foundation; either
+ version 2.1 of the License, or (at your option) any later version.
+ 
+ This library is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ Lesser General Public License for more details.
+ 
+ You should have received a copy of the GNU Lesser General Public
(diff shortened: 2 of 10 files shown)


feel/feelcore/environment.cpp
```diff
@@ -2211,6 +2211,7 @@ Environment::expand( std::string const& expr )
     boost::replace_all( res, "${datadir}", dataDir );
     boost::replace_all( res, "${exprdbdir}", exprdbDir );
     boost::replace_all( res, "${h}", std::to_string(doption("gmsh.hsize") ) );
+    boost::replace_all( res, "${testsuite_srcdir}", topSrcDir + "/testsuite/" );
```
vincentchabannes: This is not a good way because topSrcDir doesnt work in a install process and also testsuite is a subproject which can be compile out of feel++ lib.
Note : some keywords like topSrcDir will be remove of this function.

feel/feelmodels/modeloutputs.cpp
```diff
@@ -0,0 +1,134 @@
...
+    M_p( p ),
+    M_directoryLibExpr( directoryLibExpr ),
+    M_name( name )
+{
+    try {
+        M_type  = M_p.get<std::string>( "type"  );
+    }
+    catch ( pt::ptree_bad_path& e )
+    {
+        if ( Environment::isMasterRank() )
```
vincentchabannes: Use ```M_worldComm->isMasterRank()```
The Environment worldcomm mustn't be  used in feel++ lib.

feel/feelmodels/modelproperties.cpp
```diff
@@ -154,6 +155,18 @@ ModelProperties::ModelProperties( std::string const& filename, std::string const
...
+    if ( out )
+    {
+        LOG(INFO) << "Model with outputs\n";
+        if ( !directoryLibExpr.empty() )
+            M_outputs.setDirectoryLibExpr( directoryLibExpr );
+        M_outputs.setPTree( *out );
+    }
+    else
+    {
+        LOG(WARNING) << "Model does not have any outputs\n";
```
vincentchabannes: Remove this WARNING.  There is no reason to have a warning if the output section is not present.

testsuite/feelmodels/test_modelproperties.cpp
```diff
@@ -21,28 +21,32 @@
...
-
     auto mesh = loadMesh(new Mesh<Simplex<3> >);
     auto Xh = Pch<1>(mesh);
     auto g = Xh->element();
     auto d = Xh->element();
 
-    ModelProperties model_props( "test.feelpp" );
-
+    ModelProperties model_props( Environment::expand("$testsuite_srcdir/feelmodels/test.feelpp" ) );
     auto mats = model_props.materials();
```
vincentchabannes: Use an option instead hardcoded the path (because it doesn't work when testsuite  is installed)
The cfg will contain  ```myoption=$cfgdir/test.feelpp```

feel/feelmodels/modelproperties.cpp
romainhild: should I remove also the warnings for boundary/initial conditions, materials and postprocess ?

vincentchabannes: yes, you can

feel/feelmodels/modeloutputs.cpp
```diff
@@ -0,0 +1,132 @@
...
+        }
+        else
+        {
+            this->insert( std::make_pair( v.first, this->getOutput( v.second, v.first ) ) );
+        }
+    }
+}
+
+ModelOutput
+ModelOutputs::getOutput( pt::ptree const& v, std::string name )
```
prudhomm: std::string const& name

feel/feelmodels/modeloutputs.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+    std::string M_name;
+    std::string M_type;
+    std::set<std::string> M_range;
+    int M_dim;
+};
+
+std::ostream& operator<<( std::ostream& os, ModelOutput const& o );
+
+class FEELPP_EXPORT ModelOutputs: public std::map<std::string,ModelOutput>
```
prudhomm: please add documentation here

feel/feelmodels/modeloutputs.hpp
```diff
@@ -0,0 +1,95 @@
...
+#define FEELPP_MODELOUTPUTS_HPP 1
+
+#include <boost/property_tree/ptree.hpp>
+#include <feel/feelvf/ginac.hpp>
+
+namespace Feel {
+
+namespace pt =  boost::property_tree;
+
+class FEELPP_EXPORT ModelOutput
```
prudhomm: Please add documentation

prudhomm: add basic code example  
```c++
@code
<add code here>
@endcode
```

feel/feelmodels/modeloutputs.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <iostream>
+#include <boost/property_tree/json_parser.hpp>
+
+#include <feel/feelmodels/modeloutputs.hpp>
+
+namespace Feel {
+
+ModelOutput::ModelOutput( WorldComm const& worldComm )
+    :
+    M_worldComm( &worldComm )
```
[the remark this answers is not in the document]
romainhild: indeed we do that for example in modelmaterial.cpp, but in modelparameters, we do 
```
WorldComm const * M_worldComm;
```
ModelParameters::ModelParameters( WorldComm const& world )
  : M_worldComm( world )
{}
```
should I change to this ?
Or maybe remove it completely, I don't use it for now, but if we add some management of expression, we may need it later...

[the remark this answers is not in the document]
Comment (prudhomm) [reply feel/feelmodels/modeloutputs.cpp]: how about using `shared_from_this()` ? I think that at the top we use shared pointers no ?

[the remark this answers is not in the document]
Comment (prudhomm) [reply feel/feelmodels/modeloutputs.cpp]: I think that will be fine for now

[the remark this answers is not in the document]
Comment (romainhild) [reply feel/feelmodels/modeloutputs.cpp]: so `M_worldComm` should be a `boost::shared_ptr<WorldComm>` and initialized by `M_worldComm( worldComm->shared_from_this() )` ?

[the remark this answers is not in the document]
Comment (romainhild) [reply feel/feelmodels/modeloutputs.cpp]: The problem is that in modelproperties, it is not a shared pointer.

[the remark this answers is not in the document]
Comment (vincentchabannes) [reply feel/feelmodels/modeloutputs.cpp]: @prudhomm I think we need to change all WorldComm objet into shared_ptr ( and add factory methods for generate the worldComm). The shared_from_this implies some constraints, so I am not really agree for use with WorldComm class.
@romainhild  you can keep the code for now.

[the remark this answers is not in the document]
Comment (romainhild) [reply feel/feelmodels/modeloutputs.cpp]: ok I have done the other modifications.

**Comment by romainhild** [comment]
Is it ok then ?
Builkite seems to have failed but I don't have access to it, so I'm not sure if it is normal or not.
The Travis CI build passed for what it's worth...

**Comment by vincentchabannes** [comment]
@romainhild there is this error in buildkite when cmake is run :

```
CMake Error at cmake/modules/feelpp.macros.cmake:272 (file):
--
  | file COPY cannot find
  | "/home/feelpp/src/feelpp/testsuite/feelmodels/modelproperties.cfg".
  | Call Stack (most recent call first):
  | cmake/modules/feelpp.macros.cmake:357 (feelpp_add_application)
  | testsuite/feelmodels/CMakeLists.txt:27 (feelpp_add_test)
```

PR #1135, opened 2018-03-27
Feature/deim

### LINKED ISSUE #866: Support smart storage/replay of offline eim/deim data
@jbwahl @romainhild @cdaversin I have been thinking about a discussion with @jbwahl regarding storing matrices in DEIM. The issue is about having to rebuild the entire offline EIM data. storing matrices or vectors would help and avoid having to replay the entire algorithm.
However storing matrices is impractical(due to size), moreover in the case of vector and matrices we are dependent also on the partitioning.

There would be a way to achieve fast replay of EIM/DEIM by just storing the small relevant information to rebuild the vectors and matrices. the information would be the \mu_i
​​  to compute the basis (vectors or matrices or finite element fields) and the t_​i
​​  the points(and the geomap context). This should be sufficient to rebuild everything unless I am missing something but if I am missing something then we save it.

This is a good compromise because we store only very little information and we can just have an EIM/DEIM mode that takes these data and reconstruction the vectors/matrices/finite element fields
The information would be stored in a json data file as metadata for eim/deim

/cc @vincentchabannes

### LINKED ISSUE #925: Add support for DEIM
deim provides an algebraic version of eim

### LINKED ISSUE #974: Effective online phase for DEIM
- [x] Build a submesh from list of elements

- [x] Save submesh and reread it on sequential

- [x] Correspondance between dof and subspace dof

- [x] Add offline and online models in DEIM

### LINKED ISSUE #1047: DEIM : add option to store solutions on disk (NL problems)
When using multiple EIM/DEIM with NL problems, it could be interesting to store the problem solution on disk to be reread by the different algorithms.
Proposition of options : `deim.store-nl-solutions-on-disk` / `eim.store-nl-solutions-on-disk`

This will be done using the parameters as key via an hash function (for the file name). I propose to add the method `mu.key()` to the `ParameterSpace::Element` class

The directory for where the solution are written can be change via the option : `deim.store-on-disk-directory` with an common default name for EIM and DEIM.

If not written on disk the solutions are stored in memory during the greedy algorithm.

Any comments @prudhomm @romainhild @vincentchabannes

### LINKED ISSUE #1048: DEIM : add options to store Tensors during greedy
We need different option to store in memory the tensors during the greedy algorithm. 
This option has to be, by default, `true` for vector DEIM and `false` for matrix DEIM.

We have huge gain if we store all the assembled tensor in memory during greedy but this can be very costly for the matrix algorithm.

@prudhomm @romainhild

### LINKED ISSUE #1058: Deadlock in DEIM using linftyNorm for vectors
When runing DEIM test in the test suit (https://github.com/feelpp/feelpp/blob/feature/deim/testsuite/feelcrb/test_deim.cpp) we see a deadlock during the greedy algorithm of DEIM, if we use the `linftyNorm()` method on vectors.

Using another method as norm solve the problem. See https://github.com/feelpp/feelpp/blob/feature/deim/feel/feelcrb/deim.hpp#L531

DEIM is now fixed but the bug is not understood.
This bug only occurs when the option `deim.store-vectors` is set to `true`.
This bug only occurs with vectors and not with the matrix version of DEIM

@prudhomm @vincentchabannes

jbwahl:
- [x] Have you written Doxygen comments in your contribution?

-----

Last development in DEIM framework: 
- fix #1058, using `linftyNorm()`
- close #1048 : use options `--deim.store-vectors=true` and `--deim.store-matrices=true` to store tensors during DEIM/MDEIM greedy algorithm.
- close #1047 : during greedy algorithm, nl-solutions are written on disk if option `--deim.elements.write=true`. The storing directory can be specified using `--deim.elements.directory=<path>`. This directory is cleaned after greedy if options `--deim.elements.clean-directory=true`
- close  #974: we now have an interpolation model for the online phase. Also work for composite spaces, thanks to the new class : `TwoSpacesMap` which provides a map between a parallel space and a sequential space (on a submesh). see d29c7f0551ad760ec73b35789e272b2ac3c7eb85
- close #866 : DEIM can be reloaded from database. The tensors are not stored but rebuilt from saved value of mu_m.
- close #925 : DEIM is now fully operational


jbwahl:
@romainhild @vincentchabannes I am done with that feature. 
If everything is ok for both of you, we can merge.

jbwahl:
This merge will close #925 and close #974

prudhomm:
@jbwahl  could you edit the description and add the list of issues this PR is dealing with ?
thanks in advance

jbwahl:
@prudhomm is description ok for you ?

jbwahl:
@prudhomm all good here. Ready to merge.

Issue #1121, opened 2018-02-14
Support input and output data files  via Girder


prudhomm:
With MSO4SC or cloud computing, we will have often the case where the data is not available locally but remotely. In order to facilitate the use of remote data, we should be able to provide ways to 
 - retrieve test cases directories by url
 - retrieve test cases archives by url

Using CKAN and/or Girder we have ways to say where our data is, we should also think about simplifying url thanks to these metadata which will provide the true location.

this is linked to #1116

We should also be able to upload output data to Girder once simulation are done

vincentchabannes:
@prudhommit's not finished. Currently, from girder, we can download several files, but not folder or item. It require to use unzip tool.

romainhild:
Maybe it would be useful to check for errors, for example for Github, if the path is wrong or trying to access to a private repo with the wrong token, the following json is returned:
```json
{
"message": "Not Found",
"documentation_url": "https://developer.github.com/v3/repos/contents/#get-contents"
}
```
Currently, an obscur message about yamllint and jsonlint is displayed.
We could have something like the following at this [point](https://github.com/feelpp/feelpp/blob/a145810d8eb0ef060804c5f4efa34b39e7dbbd5c/feel/feelcore/remotedata.cpp#L386):
```c++
if( auto message = ptree.get_optional<std::string>("message") )
  CHECK( message != "Not Found" ) << "url is not valid or you don't have privileges to access the repo\nplease check this link http://docs.feelpp.org/toolboxes/0.105/what-is-a-toolbox/#_run_an_application";
```

what do you think @prudhomm @vincentchabannes ?

ps: I didn't find the corresponding issue for Github.

prudhomm:
Sure the obscur message  shows issues in json parsing and arrives at the very end if something went wrong, it can definitely be improved. errors can occur in `json` for the toolbox, in `json` for the download. providing in context errors is very good. however no `CHECK`!  I want to have clean exception thrown and handled later in the code so that everything is shut down properly.

vincentchabannes:
@romainhild it's not always possible to do that, only one process have this information. Each process must send to the Environment an exception type and exit properly the program.

prudhomm:
depends on the gravity of the exception. in this case that might do it, however we don't want to do it here but rather at a central place for error handling

prudhomm:
@vincentchabannes should it be closed? upload and download are now available and documented right ?

Issue #1116, opened 2018-01-15
Support test cases given by a directory 


prudhomm:
We should add a new option to the command line `--case`  that takes either directory.

This requires to have some guidelines :
 1. either one cfg and one json in directory
 2. main cfg and json files should be named after the basename of the directory 

```
$ feelpp_toolbox_csm_3d --case=sensor 
# case one (no other cfg/json file)
sensor/toto.cfg
           /toto.json
        
# case two (multiple cfg, json files) main ones given by directory basename
sensor/sensor.cfg
           /sensor.json
           /toto.cfg
           /titi.json
```

This will simplify the handling of remote test cases in MSO4SC

We could discuss also the automated support for tarball/archives and remote testcases

prudhomm:
regarding remote access support we could use `libcurl`(but a la rsync to avoid now download) and the `easy` interface to fetch the data/repository . that will be most helpful when we start working with ckan in mso4sc

PR #1164, opened 2018-05-17
Feature/remote data

vincentchabannes:

-----

Start work on remote data in Feel++ by using libcurl. Issues related is  #1116 #1121.

* implement RemoteData class which allow to download : 
** a file from an url
** a file or folderfrom github
** one or several from girder (by given the files ids)
* remote data is supported with
** ```--config-file(s)``` option (only a file)
** ```--case``` option ( only a folder) : defined the config-file of an application : if only one cfg in the folder it take this one, else it take the one with the same name of the folder. ```--case.config-file```   can be specified in order to choose the cfg file in the folder
** ```--mesh.filename``` option in the toolboxes

An example of use : 
```mpirun -np 4 ./toolboxes/thermoelectric/feelpp_toolbox_thermoelectric_3d --case="github:{path:toolboxes/thermoelectric/ElectroMagnets/HL-31_H1}" --thermo-electric.mesh.filename="girder:{file:5af862d6b0e9574027047fc8}" --case.config-file HL-31_H1_nonlinear.cfg```


prudhomm:
@vincentchabannes I approved but didn't see any test in the testsuite, I think we need to add tests in the testsuite in the future

PR #1244, opened 2018-10-23
Feature/bc markers

romainhild:
Implement #1243 and fix csm toolbox with range based function spaces

PR #1245, opened 2018-10-23
Feature/partition ofile

romainhild:
update mesh_partitioner:
- `ofile` is the name of the output file if the number of partitions is 1
- `ofile` is the prefix of the output file if the number of partitions is greater than 1
- `odir` and `ofile` can now be used at the same time


romainhild:
there is only the last two commits and the files in `feelpp/tools/mesh` which change with respect to `feature/bc-markers`

prudhomm:
@romainhild you are mixing  with PR #1244. why does it depend on #1244?

prudhomm:
you should provide an example of your use-case in the manual

romainhild:
@prudhomm it does not depend on #1244 I didn't thought of that when I made the pull request.
I can close this one, cherry pick the commits and reopen a PR, or we could merge this branch in feature/bc-markers, or wait for #1244 to be merged in develop ?

Anyway I'll add an example of what I wanted.

PR #1248, opened 2018-10-26
Feature/modelmarkers

romainhild:
@vincentchabannes @prudhomm I have created the `ModelMarkers` class, and a `ModelBoundaryConditions` class.
Is the `ModelMarkers` what you want ? I think it's ok for now, we will be able to add more features later.

For the `ModelBoundaryConditions` it could be nice to have that, I think it's a bit cleaner that the previous one. But we would need to update a lot of code, I don't know it is worth it.

It you think we could use that, we could merge this branch in bc-markers, if not we just need to cherry-pick the ModelMarkers.


romainhild:
@vincentchabannes I have made some changes here, namely adding the new classes and making feelmodels use it.
For now, `ModelBoundaryConditions` is created in `ModelProperties` along side the old `BoundaryConditions`, so it should not break any code.

Maybe I can merge this one into #1244 ?

romainhild:
I merge this PR into #1244 it will be easier not to have two PR...

PR #1259, opened 2018-12-19
Kalman filters

PhilippeRicka:
Working UKF.py python code.
Self-sufficient testcases, namely sine signal tracking and RCR indirect measurement.


PhilippeRicka:
Where should I move the examples/testcases ?

prudhomm:
@PhilippeRicka the kalman filter code is still not working properly and we need actually a set of tests showing that the code is ok.
Could you let us know where do we stand ?

/cc @mszopos

PR #1260, opened 2018-12-20
Feature/ls

metivett:
This pull request integrates the latest `feature/ls` branch with `develop`, and brings working `AdvDiffReac`, `LevelSet` and `MultiFluid` toolboxes.
It also features some refactoring of the `LevelSet` and `MultiFluid` space and tool management through the `LevelSetSpaceManager` and `LevelSetToolManager` classes, as well as optimisations regarding the reuse of advection spaces.
In addition to these core changes, the pull request also brings forth some recent developments regarding the computation of interface curvature fields and membrane forces. Note that work regarding these latter additions is still in progress.


metivett:
@vincentchabannes I know the changes with GMSH are not first-rate, but it was the quickest and most efficient I could do to be able to continue working with this branch on my laptop and on the clusters... I don't really want to have the GMSH changes on another branch, as I am actively developing and running it. Maybe we can wait for a proper fix before merging it with `develop` ?

vincentchabannes:
@metivett about Gmsh, a quick fix is to replace the line 174 of feelpp/feelpp/feel/CMakeLists.txt by this one : 
```
  target_include_directories( feelpp BEFORE PUBLIC ${GMSH_INCLUDE_DIR} )
 ```

**Comment by metivett** [comment]
@vincentchabannes I reverted the changes related to GMSH, and used the prepend include approach you mentioned ! Could you please review the PR again ? Thx :wink: !

**Comment by vincentchabannes** [comment]
thanks @metivett. I do the review asap. I have also rerun buildkite with your pr

File: toolboxes/feel/feelmodels/fluid/fluidmechanicscreate.cpp [hunk 1]
```diff
@@ -899,7 +899,7 @@ NullSpace<double> getNullSpace( SpaceType const& space, mpl::int_<3> /**/ )
 
 FLUIDMECHANICS_CLASS_TEMPLATE_DECLARATIONS
 void
-FLUIDMECHANICS_CLASS_TEMPLATE_TYPE::init( bool buildModelAlgebraicFactory )
+FLUIDMECHANICS_CLASS_TEMPLATE_TYPE::init( bool buildModelAlgebraicFactory, bool buildBlockVector )
```
vincentchabannes: @metivett why it's necessary to have this arg buildBlockVector? It changes the standard behavior of a toolbox that I assumed: after init, the block vector must be built, it's very important in multiphysics toolboxes. Also, it doesn't take a lot of memory (it's a just a vector and it's very fast)
Moreover, it breaks the python interface.

toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatejacobian.cpp
```diff
@@ -206,7 +206,7 @@ FLUIDMECHANICS_CLASS_TEMPLATE_TYPE::updateJacobian( DataUpdateJacobian & data )
 
     //--------------------------------------------------------------------------------------------------//
     // user-defined additional terms
-    this->updateJacobianAdditional( J, _BuildCstPart );
+    this->updateJacobianAdditional( data );
```
vincentchabannes: the methods updateJacobianAdditional, updateResidualAdditional, updateLinearAdditionalmethod should be removed and replaced by system proposed by the ModelAlgebraicFactory :  addFunctionJacobianAssembly, addFunctionResidualAssembly and addFunctionLinearAssembly. It's used all multiphysics toolboxes and it allows to reduce the intrusion in fluid toolbox

toolboxes/feel/feelmodels/fluid/fluidmechanicscreate.cpp
metivett: @vincentchabannes It is currently required by the `MultiFluid` initialisation, because of the possible coupling between the `FluidMechanics` and the `LevelSet` classes: in the case of an imposed surface inextensibility constraint, we need the `LevelSet`'s function-space to build the Lagrange-multiplier bloc in the algebraic vector, and we also need the `FluidMechanics` velocity space for the level-set advection.
Actually, we already discussed this kind of design when removing the two-step initialisation behaviour (with a `build` and an `init` member), which provided much more flexibility for the coupling between toolboxes, at the cost of maintainability. Maybe we can find another more robust initialisation approach ?

toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatejacobian.cpp
metivett: Ok, I am looking into this !

toolboxes/feel/feelmodels/fluid/fluidmechanicscreate.cpp
vincentchabannes: You can build the block vector in the init of fluid and upgrade this one in the init of multifluid. It’s very important to avoid intrusivity for maintains thé code. You canot change Api, the impact is big now and I want all toolboxes similar. I would also prefer to not have buildVector as virtual.

toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatejacobian.cpp
metivett: Done (c.f. f266de9) !

toolboxes/feel/feelmodels/fluid/fluidmechanicscreate.cpp
metivett: It still seems that the call to `M_blockVectorSolution.buildVector( this->backend() );` needs all the blocks to be properly constructed,  no ?
If I use the approach you suggest (build a $(U,p)$ vector in `FluidMechanics`, then "upgrade" it to add the required $L$ Lagrange-multiplier field), I can't really update the vector, I need to rebuild it completely (all the more since the `resize` function, which actually wraps the STL::vector::resize, is not guaranteed to leave the already constructed parts untouched)... Even if this is not the most costly part (and by far, I agree), this does somehow look a bit flawed, suggesting that our API is not fully robust.

vincentchabannes: @metivett I have moved `M_blockVectorSolution.buildVector( this->backend() );` into the initAlgebraicFactory ( Everything that depends on the monolithic formulation is built here).
Also, I did a change which probably breaks your code, you need to call `FM->startTimeStep()` at the beginning of the time loop else the bdf will not be started. I did that for implement Theta time stepping and a better management of initial solution

metivett: @vincentchabannes Ok, I will look into this ! Do you agree to the changes I did to remove intrusion into `FluidMechanics` ?

vincentchabannes: Completly agree

toolboxes/feel/feelmodels/levelset/levelsetadvection_inst.cpp
```diff
@@ -0,0 +1,65 @@
+
+#include "levelsetconfig.h"
```
prudhomm: use `<>` with the proper path otherwise at installation we will be in trouble

toolboxes/feel/feelmodels/multifluid/surfacetensionforcemodel.hpp
```diff
@@ -0,0 +1,101 @@
...
+           << ")";
+
+    return _ostr;
+}
+
+template<typename LevelSetType, typename FluidMechanicsType>
+void
+SurfaceTensionForceModel<LevelSetType, FluidMechanicsType>::loadParametersFromOptionsVm()
+{
+    M_surfaceTensionCoeff = doption( _name="surface-tension.coeff", _prefix=this->prefix() );
```
prudhomm: do not use option in classes directly but set them with command line in construction. We should be able to modify this kind of option from the interface. with your code we can do that only from the command line

prudhomm: this may be done elsewhere, I have checked thoroughly. This is a bad practice (that I started) which needs to be changed

vincentchabannes: @prudhomm loadParametersFromOptionsVm() is method define by all toolboxes, it's call only by the constructuor and of course we can changes the value after

feelpp/feel/CMakeLists.txt
```diff
@@ -171,7 +171,7 @@ if ( GMSH_FOUND )
   if ( FEELPP_HAS_MED )
     target_compile_definitions( feelpp PUBLIC FEELPP_HAS_MED FEELPP_HAS_GMSH_HAS_MED )
   endif()
-  target_include_directories( feelpp PUBLIC ${GMSH_INCLUDE_DIR} )
+  target_include_directories( feelpp BEFORE PUBLIC ${GMSH_INCLUDE_DIR} )
```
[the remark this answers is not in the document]
metivett: Ok, but this should then be discussed quickly, as it can break compilation, so that I need to provide help for the compilation to people who use this toolbox...

feelpp/feel/feeldiscr/operatorlagrangep1.hpp
```diff
@@ -163,7 +163,9 @@ public OperatorLinear<SpaceType, typename detailOpLagP1::SpaceToLagrangeP1Space<
                         std::string pathMeshLagP1=".",
                         std::string prefix="",
                         bool rebuild=true,
-                        bool parallelBuild=true );
+                        bool parallelBuild=true,
+                        size_type meshUpdate=MESH_RENUMBER|MESH_UPDATE_EDGES|MESH_UPDATE_FACES|MESH_CHECK
```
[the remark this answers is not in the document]
metivett: I chose this as a default case, but I am indeed not sure about it. In general, the `meshUpdate` option allows to specify the quantities to build for the Lagrange P1 mesh. This is particularly useful for high-order redistantiation of level-sets, which only require `MESH_UPDATE_FACES_MINIMAL|MESH_NO_UPDATE_MEASURES`.

feelpp/feel/feeldiscr/operatorlagrangep1.hpp
```diff
@@ -322,7 +324,9 @@ OperatorLagrangeP1<space_type>::OperatorLagrangeP1( domain_space_ptrtype const&
                                                     std::string pathMeshLagP1,
                                                     std::string prefix,
                                                     bool rebuild,
-                                                    bool parallelBuild )
+                                                    bool parallelBuild,
+                                                    size_type meshUpdate
+        )
```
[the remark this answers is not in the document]
metivett: Indeed you are :wink:, but they should !

feelpp/feel/CMakeLists.txt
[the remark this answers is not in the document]
prudhomm: Could you extract it in a small pr then?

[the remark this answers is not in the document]
prudhomm: you can leave it here and it will get merged with the small pr which will be addressed probably before

prudhomm:
also note that codacy is not too happy with your PR ;)
https://app.codacy.com/app/prudhomm/feelpp/pullRequest?prid=2770005

feelpp/feel/CMakeLists.txt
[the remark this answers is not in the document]
metivett: Yeah sure, that's also what I do. The PR is here : #1289.

metivett:
Yes, I have noticed, such a picky boy this codacy :wink:... I shall look into this as soon as possible, but it is not a top priority.

toolboxes/feel/feelmodels/levelset/levelsetadvection_inst.cpp
metivett: Actually, `levelsetconfig.h` is a file generated by CMake from `levelsetconfig.h.in`, so its directory depends on the current build (in particular the instantiation polynomial order, etc.).

toolboxes/feel/feelmodels/fluid/fluidmechanicscreate.cpp
metivett: @vincentchabannes Regarding the changes you made to `FluidMechanics`, maybe we should rename the `initAlgebraicFactory` to something more explicit if this function is supposed to build all the monolithic algebraic data (maybe something like `initAlgebraicData` ?). Moreover, we should remove the `buildBlockVector`, which now only forwards the call to `initBlockVector`...
And @vincentchabannes @prudhomm, regarding the time stepping strategy, are we planning any refactoring soon ? The interface between generic time-stepper, and the strong dependance on `BDF` API is somehow clumsy, and the toolboxes should abstract themselves from the details of the time-stepping used.

vincentchabannes: I am agree with the renaming, I didn't want to break your codes.

The bdf object is still used with theta scheme just by simplicity. A lot of time schemes (as theta scheme) are writed in the form u^{n+1} = u^n + \Delta t * F, so we can rewrite as  (u^{n+1} - u^n )/Delta t = F and we retrive the bdf order 1. This form allow to have a F which not depends of time step and avoid to have switch between bdf and other time stepper. I am agree that use of bdf should be take with caution. My next step is to implement the fractional theta step scheme (or more generaly multiple step scheme) and we will see what it's the best way to do that.

metivett: I don't exactly get your point regarding the time-stepping... Actually, it seems that if you write most of the schemes as (u^{n+1} - u^n )/Delta t = F, then F should depend on the scheme (e.g. on \theta in the theta-scheme). And there are lots of schemes which can't be written under this form (except if you put everything in F :wink:), like for example the Crank-Nicolson, which involves (u^{n+1} - u^n )/Delta t - 1/2 F^{n+1} = 1/2 F^{n}.
I think that the time-stepping classes should reflect this difference and provide a clear interface using some `timeStepBase` class, and then have multiple inheritance chains. For the particular case of the theta scheme, I think it should have its own class, allowing thereby the Crank-Nicolson special case theta=1/2.

vincentchabannes: We support now Crank-nickolson, it's a theta scheme with theta=0.5. F should depends of time n+1 but also of time n, n-1,... I do a framework for manage exactly this kind of scheme.

vincentchabannes: My works is actualy done for only for theta scheme, but it's now very easy to add others time schemes as adams-moulton

metivett: Ok, I have merged `develop` into `feature/ls` and fixed the issues related to the new `startTimeStep` of `FluidMechanics` in f0a34c6.

toolboxes/feel/feelmodels/modelcore/modelnumerical.cpp
```diff
@@ -47,6 +47,7 @@ ModelNumerical::ModelNumerical( std::string const& _theprefix, worldcomm_ptr_t c
         M_timeInitial( doption(_name="ts.time-initial") ),
         M_timeFinal( doption(_name="ts.time-final") ),
         M_timeStep( doption(_name="ts.time-step") ),
+        M_timeOrder( ioption(_prefix=_theprefix, _name="ts.order") ),
```
vincentchabannes: @metivett There is no reason to have time order here!

toolboxes/feel/feelmodels/modelcore/options.cpp
```diff
@@ -89,6 +89,8 @@ Feel::po::options_description modelnumerical_options(std::string const& prefix)
         (prefixvm(prefix,"filename").c_str(), Feel::po::value<std::string>()->default_value( "" ), "json file describing model properties" )
         (prefixvm(prefix,"mesh.filename").c_str(), Feel::po::value< std::string >(), "input mesh or geo file")
         (prefixvm(prefix,"geomap").c_str(), Feel::po::value< std::string >()->default_value("opt"), "geomap strategy : ho, opt ")
+
+        ( prefixvm( prefix, "ts.order" ).c_str(), Feel::po::value<int>()->default_value( 1 ), "time order" )
```
vincentchabannes: idem here

toolboxes/feel/feelmodels/modelcore/modelnumerical.cpp
metivett: @vincentchabannes Why not ? I think we need an "abstract" way to give the order to any model. Otherwise, the time order is only given by the BDF object, which itself reads it directly from the options, and cannot be set from the code API directly. This is a bit in conflict with the OOP approach...
In particular, I need this for `LevelSet`'s sub-advections, which must use the same time-order as the level-set advection.

metivett:
> Could you explain me what you do with near null space. I am pretty sure that I have tested that in fieldsplit (and it's crutial in solid mechanics).

Sure ! I added a way to specify the prefix of the fieldsplit you want to pass the near-null-space. This is required when using nested fieldsplits (when for example you have 3 fields: U, P and L a lagrange-multiplier ; then you do a first block with (U,P) and the second with L ; then if you want to use a fieldsplit into the (U,P) sub-fieldsplit, and attach a near-null-space for the velocity block, you can use `use-velocity-near-null-space.prefix=multifluid.fluid.fieldsplit-0`.

I agree this could be implemented in a more robust way, but the current implementation of nested fieldsplits makes this clumsy...

toolboxes/feel/feelmodels/modelcore/modelnumerical.cpp
vincentchabannes: it's too specific, you can add a virtual function which return 1 by default, but it's require to update all toolboxes. For example, with Newmark or theta scheme, the time order depends of the parameter given, the order is not just one parameter. It's your multiphysics model which should impose the compatibility of your several time scheme

metivett: Yes, I agree it's too restrictive, but I don't have a better API for the moment... In practice, I think we really need to be able to tune the time-stepping properties dynamically from within the toolboxes.

vincentchabannes: but you can,  what is the problem?

metivett: You mean, using `bdf->setOrder` ?

vincentchabannes: nope, I thought that you can store the member M_timeOrder inside the advection class. The thing that I don't like is timeOrder, I would prefer to have M_bdfTimeOrder and avoid confusion. Anyway, it's probably the time to have another base class which manages time-stepping properly.

Issue #1256, opened 2018-11-26
Compilation fails on non case sensitive system


romainhild:
On MacOS, trying to compile feelpp, we get the following error:
```
feelpp/feel/feelfilters/gmsh.cpp:66:10: warning: non-portable path to file '<context.h>'; specified path differs in case from file name on disk
      [-Wnonportable-include-path]
#include <Context.h>
```
This is due to the fact that python 3 has also a file named `context.h` and its include flag comes before the one for gmsh.

As pointed by @metivett a solution could be to use `#include <gmsh/Context.h>` and to set the include path for gmsh to `<path_to_gmsh>/include` instead of `<path_to_gmsh>/include/gmsh`.
Another solution is to filter the include flags for each file.

PR #1289, opened 2019-03-06
CMake: include GMSH directories BEFORE (prepend)

metivett:
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/feelpp/feelpp/pulls) for the same change?
- [X] Have you added an explanation of what your changes do and why you'd like us to include them?
-----

This PR provides a quick fix for the compilation issue on case insensitive OS (#1256), related to GMSH headers includes.
More advanced fixes involve more thorough analysis and accurate specifications for the includes, as the conflicting `context.h` header coming from both PYTHON and GMSH libraries should not be mixed. However solving the conflicting case (when the file requires both PYTHON's and GMSH's context.h files) is still under debate...


vincentchabannes:
this is already tested, no problem with this pr

prudhomm:
what do you mean? it is already in develop ?

vincentchabannes:
No, but I have tested (atlas, macos and docker) this one before send it the patch to @metivett .
And buildkite has already compiled the branch feature/ls without any problem.

prudhomm:
my concern is that there might be side effects with python. if you say it is good then, then let's go

vincentchabannes:
The issue is only macos with homebrew. If a problem appears we can limit this change to this architecture.
In the future, I would like to hide the gmsh include in the feel++ headers and use only in some source files. It's quite easy to do that with cmake, but it requires a recent version, so I wait

PR #1295, opened 2019-03-18
add possibility to load json partitionned mesh automatically

romainhild:
change default gmsh.filename to empty string
if filename is not empty but without extension, search for partitionned mesh filename_p<n>.json
if filename is not empty and not found, exit the application
if filename is left empty use hypercube

I think that this behavior would prevent some runtime errors when one wants to run his simulation on a given mesh but make a mistake on its name. Currently, the simulation continues on a hypercube, which I think is not what most of the users want. With this pull request the simulation would stop.
If the user does not change the filename, it will still use the hypercube, allowing quick tests without defining a geo.

Another part is to search for partitioned hdf5 mesh without specifying the number of partitions.
If a user has partitioned his mesh `cube.msh`, giving `gmsh.filename=cube` on 4 procs will look for `cube_p4.json`. It allows to have a unique config file for different number of processor.

-----

### DIFF
--- a/feelpp/feel/feelfilters/loadmesh.hpp
+++ b/feelpp/feel/feelfilters/loadmesh.hpp
@@ -117,7 +117,8 @@ BOOST_PARAMETER_FUNCTION(
             mesh_name.extension() != ".p3d" &&
             mesh_name.extension() != ".mesh" &&
             mesh_name.extension() != ".med" &&
-            mesh_name.extension() != ".arm" )
+            mesh_name.extension() != ".arm" &&
+            !mesh_name.extension().empty() )
         << "Invalid filename " << filenameExpand << " it should have either the .geo. .json or .msh extension\n";
 
 #if defined(FEELPP_HAS_GMSH_H)
@@ -253,6 +254,31 @@ BOOST_PARAMETER_FUNCTION(
             m->saveHDF5( fs::path(filenameExpand).stem().string()+".json" );
 #endif
         return m;
+    }
+
+    // search for partitionned mesh
+    if( !filenameExpand.empty() )
+    {
+#if defined(FEELPP_HAS_HDF5)
+        auto partName = filenameExpand + "_p" + std::to_string(partitions) + ".json";
+        fs::path json_fname=fs::path(Environment::findFile(partName));
+        if( fs::exists(json_fname) )
+        {
+            auto json_string = json_fname.string();
+            cout << "[loadMesh] Loading partitionned mesh in format json+h5: " << fs::system_complete(json_fname) << "\n";
+            LOG(INFO) << " Loading partitionned mesh in format json+h5: " << json_string;
+            CHECK( mesh ) << "Invalid mesh pointer to load " << json_string;
+            _mesh_ptrtype m( mesh );
+            m->setWorldComm( worldcomm );
+            m->loadHDF5( json_string );
+            return m;
+        }
+        else
+#endif
+        {
+            LOG(ERROR) << "Invalid filename " << filenameExpand << " aborting!";
+            exit(1);
+        }
     }
 #if defined( FEELPP_HAS_GMSH_H )
     mesh_name = soption(_name="gmsh.domain.shape");
--- a/feelpp/feel/options.cpp
+++ b/feelpp/feel/options.cpp
@@ -197,7 +197,7 @@ gmsh_options( std::string const& prefix )
 
     _options.add_options()
     // gmsh options
-        ( prefixvm( prefix,"gmsh.filename" ).c_str(), Feel::po::value<std::string>()->default_value( "untitled.geo" ), "Gmsh filename" )
+        ( prefixvm( prefix,"gmsh.filename" ).c_str(), Feel::po::value<std::string>()->default_value( "" ), "Gmsh filename" )
         ( prefixvm( prefix,"gmsh.depends" ).c_str(), Feel::po::value<std::string>()->default_value( "" ), "list of files separated by , or ; that are dependencies of a loaded Gmsh geometry" )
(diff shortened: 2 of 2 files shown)


feelpp/feel/feelcore/environment.cpp
```diff
@@ -1410,7 +1410,11 @@ Environment::doOptions( int argc, char** argv,
         std::reverse(configFiles.begin(),configFiles.end());
         for ( std::string const& cfgfile : configFiles )
         {
-            if ( !fs::exists( cfgfile ) ) continue;
+            if ( !fs::exists( cfgfile ) )
+            {
```
vincentchabannes: not good because there is by default config-file name. So all app which not requires a cfg will print a warning (a lot of app of the testsuite for example).

vincentchabannes: I do not really fan about to have default cfg. I think also options config-file and config-files should be gathered.
In this case, we don't manipulate a string but a vector of string. I think it's cleaner to have an empty vector than an empty string.

romainhild: The default config-file name should be empty, no ? So we know that if there is one or more config files, it's the user that entered them and they should be valid.
We could check for `appname+".cfg"` if no config file is found to keep the compatibility with previous versions.

prudhomm: how about adding an option that check/ensure the existence of the configfile. by default it would be off (to follow current behavior) but would allow user to crash the app if the config files are not existing/found ?

romainhild: I think that the solution I proposed should not break existing codes.
- If a user do not change the config file name, it checks the `appname.cfg` and if not found continue without configuration (as it happens currently).
- If a user gives a config file name, this file should exists otherwise the app gives a clear warning, or even crash.

@Trophime had this kind of error today, the config file was not read and there is no clear indication of that in the logs or in the output.

Trophime: I'm fine with @prudhomm proposition. 
That could be used as a way to check the consistency of cfg and more generally input data
which is **REALLY** needed

vincentchabannes: @prudhomm I think the current behavior (search a default cfg or mesh) is "contre intuitif".  It's a feeling shared by some users. We can add an option (--demo for example) which allow to use a default cfg.

prudhomm: you mean if the application is called `a` we look for `a.cfg` is that it ?

vincentchabannes: yes

prudhomm: but it is done only if no cfg is provided, no ?

prudhomm: maybe the mesh is more problematic and can do weird stuff. It can easily be disabled

vincentchabannes: the idea is to merge config-file and config-files

romainhild:
I have changed the `config-file` to a vector of string, so we shouldn't need `config-files` anymore (but I left it).
What is done:
- if no config file is given (either by `config-file`, `config-files` or `case` option) by the user, it search for `appname.cfg`, if found it adds it
- if a file is given by the user and not found, it gives a warning
- if no config file is given by the user and `appname.cfg` does not exist, we do nothing

For the mesh, I have updated the testsuite, there was some tests that used `gmsh.filename` with dummy name (toto, mygeo, hdiv,...) I remove the option of the config file and now it works.
But I think that
```
            LOG(ERROR) << "Invalid filename " << filenameExpand << " aborting!";
            exit(1);
```
is not a correct way to exiting a parallel application. What should be done ?

I have launched the testsuite and 3 tests failed, `test_element` with the weird error where it complains that an error at 1e-15 is not good enough for a tolerance of 1e-13, and `test_eim_grepl` where it complains during the serialization that the version is unsupported, I looked and it seems that it is because it is trying to load something serialized with a more recent version of boost. I used 1.65, do somebody tried with more than that ?

vincentchabannes:
@romainhild I do not totally agree with your changes. Some applications (as mesh_partitioner) are run without cfg file, only command line options. In this case, you will have a warning and I don't want to see that. It's why I proposed to add an option like --demo which tries to run an app with app.cfg.

Also, if the cfg file given by a user does not exist, It will better to stop the application (it avoid to have a crash after with a big and incompressible msg.

@romainhild We need to discuss and decide  before doing the changes, it's not a big priority I think

romainhild:
If a user does not give a config file, there is no warning:
>if no config file is given by the user and appname.cfg does not exist, we do nothing

I agree that we should quit with a message, but I think that the way I've done it for the mesh is not correct, especially in parallel.

prudhomm:
One of the issue with some warning/error handling we have is that we are trying to make a decision (eg like exiting) at a place where we don't have enough information to make a proper decision.
We should throw an exception and let the one who catches it make the (better) decision.
In general we should avoid exiting and use throw to let the upper level do their work and decice.

vincentchabannes:
@romainhild If you have an app.cfg but you don't want to use it but use only command line option, you can't or you not do really what defined exactly.
I repeat sorry but --demo  is really better and expressive for a user which don't know the feel++ machinery.

romainhild:
Yes indeed I see your point, it could lead to unwanted behavior! So we should just add an option `demo` and we try to load `appname.cfg` only if this true ?

romainhild:
> One of the issue with some warning/error handling we have is that we are trying to make a decision (eg like exiting) at a place where we don't have enough information to make a proper decision.
> We should throw an exception and let the one who catches it make the (better) decision.
> In general, we should avoid exiting and use throw to let the upper level do their work and decide.

I agree with you that it should be done like this in general, and for example for the config file, but here, I think that for `loadMesh` we have all the informations needed. And I don't think that it is a good idea to have to surround `loadMesh` with try and catch, it is usually used in the main.

PR #1307, opened 2019-03-29
Fix: wrong use of isStationaryModel instead of isStationary to test f…

### LINKED ISSUE #1306: FluidMechanics toolbox crashes with Stokes model
This problem comes from confusion between `isStationary` and `isStationaryModel` in `updateLinearPDE` (c.f. https://github.com/feelpp/feelpp/blob/develop/toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatelinear.cpp#L61), leading to the request of non-provided double at https://github.com/feelpp/feelpp/blob/develop/toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatelinearbc.cpp#L37.

metivett:
…or time-stepping.scaling addDoubleInfo (fixes #1306)

-----

### DIFF
--- a/toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatelinear.cpp
+++ b/toolboxes/feel/feelmodels/fluid/fluidmechanicsupdatelinear.cpp
@@ -58,7 +58,7 @@ FLUIDMECHANICS_CLASS_TEMPLATE_TYPE::updateLinearPDE( DataUpdateLinear & data ) c
     this->timerTool("Solve").start();
 
     double timeSteppingScaling = 1.;
-    if ( !this->isStationaryModel() )
+    if ( !this->isStationary() )
     {
         if ( M_timeStepping == "Theta" )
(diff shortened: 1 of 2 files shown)


metivett:
Actually, I don't understand your point either :wink:, but I do have a crash coming from a failed check at `F0329 15:52:35.756933 2991552 modelalgebraic.hpp:75] Check failed: this->hasDoubleInfo( info ) double info fluid.time-stepping.scalingis missing` when running for example a fluid testcase (TurekHron/cfd3.cfg) with Stokes.

vincentchabannes:
I forgot the Stokes case, I will fix that

metivett:
I still think that the `isStationary` and `isStationaryModel` tests should not be fixed (you use `isStationaryModel` to check wether we need to add the `time-stepping.scaling` in `updateLinearPDE`, but you then use the `isStationary` test to access this supposedly provided `time-stepping.scaling`.

PR #1329, opened 2019-05-24
Merge crb branches and implement empirical quadrature

romainhild:

-----

Merge the branch crbaero and biotsavart:
- use of crbblock for trilinear, saddlepoint, and aero problems
- update deim
- create biotsavart classes in mor/thermoelectric

Implement the empirical quadrature:
- add class `expressionevaluator` in feelvf to handle generically evaluation of expression regardless of the type of the expression
- add class 'empiricalquadrature`in feelcrb to create the linear programming problem, solve it and store the new quadrature points and weights
- add class to interface glpk library in feelopt


prudhomm:
please have a look at the codacy complaints. I will have a look at the recent commits. are there any tests. there doesn't seem to be any test in the testsuite, is that correct ?

romainhild:
Indeed there is no test for the empirical quadrature. I'm working on it, along with the codacity complaints.

prudhomm:
add very simple test first then increase complexity

romainhild:
@prudhomm regarding codacity, the remaining errors are not really errors...
Some are members of class not initialized in the constructor whereas they are in the other constructor that is called, or members that do not exists anymore...
And the others are a confusion between the `exp` function from the standard library and the feelpp one.

romainhild:
@prudhomm it should be ok now.
The only tests that failed on atlas were `operatorinterpolation` but @vincentchabannes is on it, and `element` where there is still this error
```error: in "element/test_element_1": difference{1.55431e-15} between v->dot( w ){1.0000000000000016} and 1{1} exceeds 1.5e-13%```
where the error is less than the tolerance but it still fails...

Issue #1328, opened 2019-05-22
Support View in CreateSubmesh


vincentchabannes:
When a createSubmesh is applied, data structures of the new mesh is completely rebuilt. In the case where we apply a moving mesh on the parent mesh, we also need to apply a moving mesh in the submesh (and require to interpolate also the displacement on the submesh).

It will be useful when we create a submesh to have an option to say that the nodes as a view of the parent mesh. The moving mesh will be applied automatically and can also reduce memory footprint.

A solution is to share the node vector (actually boost ublas node) with a shared_ptr (or directly with a pointer but the memory management will be more difficult.
@prudhomm What do you think?

Issue #1317, opened 2019-04-18
Reduce memory footprint in Mesh and FunctionSpace related data structures


prudhomm:
Use 32-bits integer rather than 64 bits integers in mesh and function space data structures to lower significantly the memory footprint

/cc @romainhild @Trophime

prudhomm:
first working results using 32bits unsigned int for indices rather than 64bits:
```
ctest -R feelpp_qs_laplacian_2d 
Test project /home/prudhomm/scratch/feelpp/hdg.clang.debug/feelpp/quickstart
      Start  1: feelpp_qs_laplacian_2d-circle-dirichlet-np-6
 1/18 Test  #1: feelpp_qs_laplacian_2d-circle-dirichlet-np-6 ..........   Passed    1.46 sec
      Start  2: feelpp_qs_laplacian_2d-circle-dirichlet-np-1
 2/18 Test  #2: feelpp_qs_laplacian_2d-circle-dirichlet-np-1 ..........   Passed    1.44 sec
      Start  3: feelpp_qs_laplacian_2d-circle-robin-np-6
 3/18 Test  #3: feelpp_qs_laplacian_2d-circle-robin-np-6 ..............   Passed    1.45 sec
      Start  4: feelpp_qs_laplacian_2d-circle-robin-np-1
 4/18 Test  #4: feelpp_qs_laplacian_2d-circle-robin-np-1 ..............   Passed    1.97 sec
      Start  5: feelpp_qs_laplacian_2d-circle-all-np-6
 5/18 Test  #5: feelpp_qs_laplacian_2d-circle-all-np-6 ................   Passed    1.55 sec
      Start  6: feelpp_qs_laplacian_2d-circle-all-np-1
 6/18 Test  #6: feelpp_qs_laplacian_2d-circle-all-np-1 ................   Passed    1.68 sec
      Start  7: feelpp_qs_laplacian_2d-triangle-oneelt-1-np-6
 7/18 Test  #7: feelpp_qs_laplacian_2d-triangle-oneelt-1-np-6 .........   Passed    1.21 sec
      Start  8: feelpp_qs_laplacian_2d-triangle-oneelt-1-np-1
 8/18 Test  #8: feelpp_qs_laplacian_2d-triangle-oneelt-1-np-1 .........   Passed    0.93 sec
      Start  9: feelpp_qs_laplacian_2d-triangle-oneelt-x+y-np-6
[...]

PR #1337, opened 2019-06-27
Feature/submesh view fsi refactoring

vincentchabannes:

-----

- implement #1328
- reduce memory footprint in mesh #1317 : measures are only stored for elements entity, other sub-entities compute measure on the fly.
- use submesh view in the fluid and fsi toolboxes: simplify the management when the mesh moving is applied
- createSubmesh use now boost parameters arguments 
- on(..) keyword for nodal projection take another arg close which allows to sync the values at interprocess


vincentchabannes:
@prudhomm it's ready for the merge. The codacy issues are not a problem, because it doesn't take into account constructor delegation.

prudhomm:
The testsuite with the new tests run properly ?

vincentchabannes:
yes, see [here](https://buildkite.com/feelpp/feelpp-testsuite/builds/237)

vincentchabannes:
@prudhomm review?

feelpp/feel/feeldiscr/createsubmesh.hpp
```diff
@@ -393,7 +400,8 @@ CreateSubmeshTool<MeshType,IteratorRange,TheTag>::build( mpl::int_<MESH_ELEMENTS
             } // for (unsigned int n=0 ... )
 
             // Add an equivalent element type to the new_mesh
-            auto const& e = newMesh->addElement( newElem,true );
+            auto eit = newMesh->addElement( newElem,true );
```
prudhomm: would it be possible to extract it here the iterator and the boolean ?

feelpp/feel/feeldiscr/createsubmesh.hpp
```diff
@@ -393,7 +400,8 @@ CreateSubmeshTool<MeshType,IteratorRange,TheTag>::build( mpl::int_<MESH_ELEMENTS
...
+            auto const& e = eit.first->second;
```
prudhomm: and here extract explicit the different members of the pair so that it is more explicit?
that way the code is more readable.

those two lines are everywhere and I(and probably others) need to remember what is the content of the pair/tuple.
may be we could query/replace the code automatically ?

feelpp/feel/feeldiscr/createsubmesh.hpp
prudhomm: we would have
```c++
auto [eit,inserted] = newMesh->addElement( newElem,true );
auto const& [key,e] = *eit;
```

prudhomm: perhaps a good regexp could automate this?
all these lines are pretty much the same

feelpp/feel/feeldiscr/createsubmesh.hpp
vincentchabannes: @prudhomm today, I have started the changes that you propose with addElement. an automatically way is quite dangerous.

prudhomm: ok. so by hand?

prudhomm:
@vincentchabannes with the `createSubmesh` in view mode, getting `boundaryfaces(submesh)` would work? 
without creating a submesh we would not be able to get the boundaryfaces right?
I haven't found tests on this

feelpp/feel/feeldiscr/createsubmesh.hpp
vincentchabannes: Yes, there are not so many. I compile currently the testsuite

vincentchabannes:
in the view mode, only the nodes (the ublas vector) are shared, thus it doesn't change the behavior of any filters

vincentchabannes:
and you can get the boundary faces from a range  of element by using the MeshSupport class

Issue #1359, opened 2019-08-26
Support create linear forms from bilinear forms 


prudhomm:
when multiplying a trial vector to the matrix representation of a bilinear form we get a vector representing a linear form in the test space of the bilinear form.
we should provide an interface for this, it can allow some optimisation when repeatedly applying data to a linear form

Issue #1358, opened 2019-08-26
Optimize TimeSet


prudhomm:
Exporters use Timeset data structure to store mesh and fields. In some case optimisation can be obtained when the fields are P1/Q1 and are defined on the same mesh as the exporter. In that case no need to create assocciated function spaces

prudhomm:
We use `constexpr` to optimize `interpolate` when the spaces are the same

Issue #1357, opened 2019-08-26
Optimize ExporterEnsight 


prudhomm:
The save step in exporter ensight is quite slow, this needs to be investigated. An issue is that we compute over and over certain data structure to transfer from feelpp to ensight

Issue #1356, opened 2019-08-26
Optimize integration on MeshStructured


prudhomm:
When using MeshStructured and in particular cartesian grids, we have optimisation opportunities for integration steps: the cells are all the same

PR #1360, opened 2019-08-26
Optimize/holo3

prudhomm:
- [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/feelpp/feelpp/pulls) for the same change?

-----
Implements #1356 #1357 #1358 and #1359 

implement also #831 hbf parallel support

### DIFF
--- a/feelpp/feel/feeldiscr/timeset.hpp
+++ b/feelpp/feel/feeldiscr/timeset.hpp
@@ -670,7 +670,7 @@ public:
         FEELPP_NO_EXPORT void add( std::string const& __n, std::string const& __fname, FunctionType const& func, mpl::bool_<false>, mpl::bool_<true>,
                                    typename std::enable_if<FunctionType::is_scalar>::type* = nullptr)
         {
-            bool extendeddof = (soption(_name="exporter.format") == "ensightgold");
+            bool extendeddof = false;//(soption(_name="exporter.format") == "ensightgold");
             boost::timer t;
 
                 if ( !M_ts->M_scalar_p1 )
@@ -747,7 +747,7 @@ public:
         FEELPP_NO_EXPORT void add( std::string const& __n, std::string const& __fname, FunctionType const& func, mpl::bool_<false>, mpl::bool_<true>,
                                    typename std::enable_if<FunctionType::is_vectorial>::type* = nullptr)
             {
-                bool extendeddof = (soption(_name="exporter.format") == "ensightgold");
+                bool extendeddof = false; //(soption(_name="exporter.format") == "ensightgold");
                 boost::timer t;
 
                 if ( !M_ts->M_vector_p1 )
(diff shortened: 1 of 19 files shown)


prudhomm:
@vincentchabannes these commits optimize things quite a bit thanks mainly to `if constexpr` and caching. let me know what you think

feelpp/feel/feeldiscr/timeset.hpp
```diff
@@ -679,9 +682,21 @@ class TimeSet
                     }
                     else
                     {
-                        M_ts->M_scalar_p1 = scalar_p1_space_type::New(_mesh=M_mesh.get(),
-                                                                      _worldscomm=func.worldsComm(),
-                                                                      _extended_doftable=std::vector<bool>(1,extendeddof) );
+                        if constexpr ( std::is_same_v<scalar_p1_space_type,typename FunctionType::functionspace_type> )
+                        {
+                            if ( ( func.mesh() == M_mesh ) && ( func.functionSpace()->basisOrder()[0] == 1 ) )
```
vincentchabannes: @prudhomm in fact, we cannot apply this optimisation generally for the ensightgoldformat because it requires to have the extended dof table. Another check is necessary : 
```
if ( ... &&  func.functionSpace().extendedDofTable() == extendeddof )
```

feelpp/feel/feeldiscr/timeset.hpp
```diff
@@ -707,7 +722,8 @@ class TimeSet
                 if ( func.worldComm().isActive() )
                 {
                     // interpolate field on visualisation space
-                    interpolate( M_scalar_p1, func, M_nodal_scalar[__fname] );
+                    interpolate( M_scalar_p1, func, M_nodal_scalar[__fname], INTERPOLATE_DIFFERENT_MESH );
```
vincentchabannes: why there is INTERPOLATE_DIFFERENT_MESH ?

prudhomm: it is not used anymore, it should be removed

feelpp/feel/feeldiscr/timeset.hpp
prudhomm: yes indeed :(

I am wondering  : is it indeed really necessary to use the extended dof in ensight ? we still have those interfaces we see in 3D

vincentchabannes: if I remember, the extended doftable removes interface with Ensight software but not in Paraview.
Anyway, I do not fully agree to have this kind of space here, and I am sure that we can retrieve these ghosts information quite easily. Maybe we can remove these feature by default and try to re-add in the future  the  ghost information without the extended doftable

prudhomm:
@vincentchabannes please review. 
I removed the extended dof for the ensight exporter which allows to optimize a bit the ensight exporter

prudhomm:
All good on the testsuite front @vincentchabannes

prudhomm:
implement also #831 hbf parallel support

PR #1370, opened 2019-10-06
Compilation failing on macOS with clang++

prj-:

-----
There are still many compilation failures, e.g.,
```
feelpp/feelpp/feel/../feel/feeldiscr/functionspace.hpp:4658:53: note: passing argument to parameter 'worldscomm' here
                                 worldscomm_ptr_t & worldscomm = Environment::worldsComm(nSpaces),
                                                    ^
feelpp/feelpp/feel/../feel/feeldiscr/functionspace.hpp:4653:56: error: binding value of type 'const vector<...>' to reference to
      type 'vector<...>' drops 'const' qualifier
        return NewImpl( mesh, cms.M_meshSupportVector, worldscomm, components, periodicity, edt );
```
or
```
feelpp/feelpp/feel/../feel/feelvf/projectors.hpp:265:1: note: candidate template ignored: could not match 'Expr' against 'tuple'
project_impl( std::shared_ptr<FunctionSpaceType> const& __functionspace,
^
feelpp/feelpp/feel/../feel/feelvf/projectors.hpp:424:12: error: no matching function for call to 'project_impl'
    return project_impl( space, range, expr, geomap );
```
Feel++ is not supposed to be built on macOS with clang++?

### DIFF
--- a/feelpp/feel/feelalg/solvereigen.hpp
+++ b/feelpp/feel/feelalg/solvereigen.hpp
@@ -540,7 +540,7 @@ protected:
  *
  * \return eigen modes
  */
-BOOST_PARAMETER_MEMBER_FUNCTION( ( typename SolverEigen<double>::eigenmodes_type ),
+BOOST_PARAMETER_FUNCTION( ( typename SolverEigen<double>::eigenmodes_type ),
                                  eigs,
                                  tag,
                                  ( required
@@ -617,7 +617,7 @@ struct compute_eigs_return_type
     typedef std::vector<std::pair<value_type,typename A_type::element_2_type> > type;
 };
 
-BOOST_PARAMETER_MEMBER_FUNCTION( ( typename compute_eigs_return_type<Args>::type ),
+BOOST_PARAMETER_FUNCTION( ( typename compute_eigs_return_type<Args>::type ),
                                  veigs,
                                  tag,
                                  ( required
--- a/feelpp/feel/feelcore/environment.hpp
+++ b/feelpp/feel/feelcore/environment.hpp
@@ -656,8 +656,8 @@ public:
         ( required
           ( name,( std::string ) ) )
         ( optional
-          ( sub,( std::string ),"" )
-          ( prefix,( std::string ),"" )
+          ( sub,( std::string ),std::string() )
+          ( prefix,( std::string ),std::string() )
           ( vm, ( po::variables_map const& ), Environment::vm() )
         ) )
     {
--- a/feelpp/feel/feelalg/functionspetsc.cpp
+++ b/feelpp/feel/feelalg/functionspetsc.cpp
@@ -292,9 +292,13 @@ PetscConvertKSPReasonToString( KSPConvergedReason reason )
 #endif
     case KSP_DIVERGED_INDEFINITE_MAT : return "DIVERGED_INDEFINITE_MAT";
 #if PETSC_VERSION_GREATER_OR_EQUAL_THAN( 3, 6, 0 )
+#if !PETSC_VERSION_GREATER_OR_EQUAL_THAN( 3, 12, 0 )
     case KSP_DIVERGED_PCSETUP_FAILED : return "DIVERGED_PCSETUP_FAILED";
+#else
+    case KSP_DIVERGED_PC_FAILED      : return "DIVERGED_PC_FAILED";
 #endif
-    case KSP_CONVERGED_ITERATING : return "CONVERGED_ITERATING";
+#endif
+    case KSP_CONVERGED_ITERATING     : return "CONVERGED_ITERATING";
 
     default: return "INDEFINE_KSP_REASON";
 
@@ -313,7 +317,9 @@ PetscConvertSNESReasonToString( SNESConvergedReason reason )
     case SNES_CONVERGED_SNORM_RELATIVE : return "CONVERGED_SNORM_RELATIVE";// =  4, /* Newton computed step size small; || delta x || < stol */
 #endif
     case SNES_CONVERGED_ITS            : return "CONVERGED_ITS";           // =  5, /* maximum iterations reached */
+#if !PETSC_VERSION_GREATER_OR_EQUAL_THAN( 3, 12, 0 )
     case SNES_CONVERGED_TR_DELTA       : return "CONVERGED_TR_DELTA";      // =  7,
+#endif
         /* diverged */
     case SNES_DIVERGED_FUNCTION_DOMAIN : return "DIVERGED_FUNCTION_DOMAIN";// = -1, /* the new x location passed the function is not in the domain of F */
     case SNES_DIVERGED_FUNCTION_COUNT  : return "DIVERGED_FUNCTION_COUNT"; // = -2,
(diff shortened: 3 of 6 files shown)


vincentchabannes:
yes it should compile on MacOS. I have compiled some feel++ app few week ago (using /usr/bin/clang++ compiler)

prj-:
OK, good to know. I think I provided enough evidences to show you that it doesn't anymore. I've fixed what I could, but we are not quite there yet... Also, you are using the deprecated `boost/timer.hpp` which should be replaced by `boost/timer/timer.hpp`, but that is less important than having a compilable library, obviously.

prudhomm:
@prj- what is your exact setup on macosx ? 
 - Os X version
 - clang version
 - Boost version 
 
PETSc is obviously 3.12

prudhomm:
@prj-  if you use homebrew, it uses boost 1.71. I am not sure that we ported Feel++ to Boost 1.71. At least I never tested, @vincentchabannes?. It might explain some changes you had to do

prj-:
OSX 10.14.6, PETSc master, SLEPc master, Gmsh master, Boost 1.71, no HDF5, clang-1100.0.33.8, MPICH 3.3.1, Eigen 3.3.7.

vincentchabannes:
for info : boost 1.69 and 1.70 are bugged with boost mpi. I have patched this one but appears only from boost 1.71 (that I never tested).

vincentchabannes:
and you have [here](http://docs.feelpp.org/user/0.107/install/prerequisites-dev/) the dependencies with versions that (we have tested)! If you use another version we can't be sure that everything is fine!

prj-:
1. most of the fixes I've provided are not related to specific dependencies, e.g., using `BOOST_PARAMETER_MEMBER_FUNCTION` outside the body of a class it not valid even with [1.67](https://www.boost.org/doc/libs/1_67_0/libs/parameter/doc/html/index.html#parameter-enabled-member-functions), using `""` (`const char[1]`) as a default value for an `std::string` should be replaced by the default `std::string` constructor `std::string()`
2. please let me know when this builds with recent dependencies, as I can't downgrade my system. Disclaimer: I still believe the remaining errors are not related to any dependencies, but instead ill-formed C++, as detailed on point 1.

feelpp/feel/feelcore/environment.hpp
```diff
@@ -656,8 +656,8 @@ class FEELPP_EXPORT Environment
         ( required
           ( name,( std::string ) ) )
         ( optional
-          ( sub,( std::string ),"" )
```
prudhomm: we never had any issue with this. I don't see why we cannot do this, there is a constructor with `const char*`
https://en.cppreference.com/w/cpp/string/basic_string/basic_string

what is the error message exactly ?

feelpp/feel/feelalg/solvereigen.hpp
```diff
@@ -540,7 +540,7 @@ class SolverEigen
  *
  * \return eigen modes
  */
-BOOST_PARAMETER_MEMBER_FUNCTION( ( typename SolverEigen<double>::eigenmodes_type ),
+BOOST_PARAMETER_FUNCTION( ( typename SolverEigen<double>::eigenmodes_type ),
```
prudhomm: ok thanks, never had an issue before. glad you fixed it

feelpp/feel/feelcore/environment.hpp
prj-: How do you take the `.size()` of a `const char*`?
```
feelpp/feelpp/feel/../feel/feelcore/environment.hpp:669:18: error: member reference base type
      'char const[1]' is not a structure or union
        if ( !sub.empty() )
              ~~~^~~~~~
feelpp/feelpp/feel/../feel/feelcore/environment.hpp:654:5: note: in instantiation of function
      template specialization
```

prudhomm:
The type for sub is std::string. The type is given as second parameter. Hence we should be able to get size() and empty()

prudhomm:
"" is just the initialisation /construction for the std::string

prj-:
This implies that Boost will always cast the default value to the given type. Is this true for all Boost versions? Given the compilation error, I'd say, no :)

prudhomm:
this is new with Boost 1.71.  Very strange that suddenly the behavior changes. Anyway no problem with an empty string of course

prj-:
Thanks, if you get a chance to investigate the error
```
feelpp/feelpp/feel/../feel/feelvf/projectors.hpp:265:1: note: candidate template ignored: could not match 'Expr' against 'tuple'
project_impl( std::shared_ptr<FunctionSpaceType> const& __functionspace,
^
feelpp/feelpp/feel/../feel/feelvf/projectors.hpp:424:12: error: no matching function for call to 'project_impl'
    return project_impl( space, range, expr, geomap );
```
That'd be great.

prudhomm:
could you give some more context? what file do you compile? how do I reproduce this ?

prj-:
Here is a more complete log. Same configuration as before. I don't think this has anything to do with Gmsh, PETSc, SLEPc, or HDF5. So maybe just try to bump your Boost version to reproduce this?
```
[ 73%] Building CXX object feelpp/feel/CMakeFiles/feelpp.dir/feells/reinit_fms_inst_2d_g2.cpp.o
In file included from feelpp/feel/feells/reinit_fms_inst_2d_g2.cpp:24:
In file included from feelpp/feel/../feel/feells/reinit_fms_impl.hpp:30:
In file included from feelpp/feel/../feel/feells/reinit_fms.hpp:40:
In file included from feelpp/feel/../feel/feelalg/backend.hpp:33:
/usr/local/include/boost/timer.hpp:21:1: warning: This header is deprecated. Use the facilities in
      <boost/timer/timer.hpp> instead. [-W#pragma-messages]
BOOST_HEADER_DEPRECATED( "the facilities in <boost/timer/timer.hpp>" )
^
/usr/local/include/boost/config/header_deprecated.hpp:23:37: note: expanded from macro 'BOOST_HEADER_DEPRECATED'
# define BOOST_HEADER_DEPRECATED(a) BOOST_PRAGMA_MESSAGE("This header is deprecated. Use " a " instead.")
                                    ^
/usr/local/include/boost/config/pragma_message.hpp:24:34: note: expanded from macro 'BOOST_PRAGMA_MESSAGE'
# define BOOST_PRAGMA_MESSAGE(x) _Pragma(BOOST_STRINGIZE(message(x)))
                                 ^
<scratch space>:148:2: note: expanded from here
 message("This header is deprecated. Use " "the facilities in <boost/timer/timer.hpp>" " instead.")
 ^
In file included from feelpp/feel/feells/reinit_fms_inst_2d.cpp:24:
In file included from feelpp/feel/../feel/feells/reinit_fms_impl.hpp:33:
feelpp/feel/../feel/feelvf/projectors.hpp:424:12: error: no matching function for call
      to 'project_impl'
    return project_impl( space, range, expr, geomap );
```

prudhomm:
indeed it seems like a boost 1.71 issue. I will have a look

prudhomm:
@prj- there have been many changes in boost.parameter/1.71 and they broke quite a bit our usage of it.
I am trying to pin down the issues. The document for boost.parameter 1.71 is a bit buggy at boost.org, it doesn't help

prudhomm:
my understanding is that something changed with requirements/constraints and our code fails to compile

Issue #1405, opened 2019-12-15
Upgrade Eigen 


prudhomm:
eigen has not been updated for some time. fixes and new features are available

prudhomm:
here is the result of testsuite

```
99% tests passed, 1 tests failed out of 394                                                                                                                            
                                                                                                                                                                       
Label Time Summary:                                                                                                                                                    
testalg              =  47.02 sec*proc (24 tests)                                                                                                                      
testcore             =  82.17 sec*proc (50 tests)                                                                                                                      
testcrb              = 152.81 sec*proc (17 tests)                                                                                                                      
testdiscr            = 452.48 sec*proc (112 tests)                                                                                                                     
testevents           =   3.75 sec*proc (2 tests)                                                                                                                       
testfilters          =  57.39 sec*proc (16 tests)                                                                                                                      
testfit              =   7.17 sec*proc (4 tests)                                                                                                                       
testintegration      =  43.48 sec*proc (16 tests)                                                                                                                      
testinterpolation    =  78.88 sec*proc (23 tests)                                                                                                                      
testleaks            =  31.53 sec*proc (16 tests)                                                                                                                      
testls               =   5.43 sec*proc (3 tests)                                                                                                                       
testmaterial         =   2.58 sec*proc (2 tests)                                                                                                                       
testmath             =   2.93 sec*proc (2 tests)                                                                                                                       
testmesh             =  64.52 sec*proc (26 tests)                                                                                                                      
testmodels           =   3.56 sec*proc (2 tests)                                                                                                                       
testopt              =   3.81 sec*proc (2 tests)                                                                                                                       
testpde              =  18.47 sec*proc (4 tests)                                                                                                                       
testpoly             =  51.47 sec*proc (20 tests)                                                                                                                      
testpython           =  12.55 sec*proc (6 tests)                                                                                                                       
testts               =  12.97 sec*proc (7 tests)                                                                                                                       
testvf               = 127.46 sec*proc (40 tests)                                                                                                                      
                                                                                                                                                                       
Total Test time (real) = 1267.67 sec                                                                                                                                   
                                                                                                                                                                       
The following tests did not run:                                                                                                                                       
        321 - feelpp_test_bdf3-np-1 (Skipped)                                                                                                                          
        332 - feelpp_test_laplacianv-np-6 (Skipped)                                                                                                                    
        333 - feelpp_test_laplacianv-np-1 (Skipped)                                                                                                                    
[...]

PR #1407, opened 2019-12-15
update eigen to git from 20191213

prudhomm:
- [X] Have you added an explanation of what your changes do and why you'd like us to include them?
- [X] Have you successfully run the Feel++ testsuite with your changes locally?

-----
See #1405  for the details


prudhomm:
sure, it is planned anyway, eigen3 has moved now to gitlab which makes it possible to do this kin off things

as to the version, #1405 shows master is good but we can try to be on the safe side and use the version  ;)

vincentchabannes:
but the testsuite is not (yet) perfect. For example, my issue with eigen value problem is not detected.

prudhomm:
sure no problem

Issue #1296, opened 2019-03-20
Support boost 1.69.0-1.70.0


vincentchabannes:
Require some fixes : 
- [x] Install of boost add an archicteture prefix (x64 ), detection require a recent >=3.12 cmake and we must define cmake variable Boost_ARCHITECTURE
- [x] Lambda expression or functor can't be pass in boost mpi argument ( all arg/parameters of a functor should be static) . This was used in feel/feells/reinit_fms_impl.hpp
- [x] mpi serialization with std::vector fails when mpi::wait_all() is called

vincentchabannes:
I have found a bug in boost mpi, I have send a fix with this pull request https://github.com/boostorg/mpi/pull/84 . I hope it will be accepted before the 1.70 release.
It means that we need to patch  the release 1.69 of boost.

vincentchabannes:
unfortunately, boost mpi has still some issues. quite hard to reproduce because not appears all the times.
```
terminate called after throwing an instance of 'boost::wrapexcept<boost::mpi::exception>'
  what():  MPI_Test: MPI_ERR_TRUNCATE: message truncated
*** Aborted at 1576576705 (unix time) try "date -d @1576576705" if you are using GNU date ***
terminate called after throwing an instance of 'boost::wrapexcept<boost::mpi::exception>'
  what():  MPI_Test: MPI_ERR_TRUNCATE: message truncated
*** Aborted at 1576576705 (unix time) try "date -d @1576576705" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x5c00000386c) received by PID 14444 (TID 0x7fa9aba2b500) from PID 14444; stack trace: ***
terminate called after throwing an instance of 'boost::wrapexcept<boost::mpi::exception>'
terminate called after throwing an instance of '    @     0x7fa973909390 (unknown)
boost::wrapexcept<boost::mpi::exception>'
  what():  MPI_Test: MPI_ERR_TRUNCATE: message truncated  what():  MPI_Test: MPI_ERR_TRUNCATE: message truncated

*** Aborted at 1576576705 (unix time) try "date -d @1576576705" if you are using GNU date ***
*** Aborted at 1576576705 (unix time) try "date -d @1576576705" if you are using GNU date ***
PC: @                0x0 (unknown)
*** SIGABRT (@0x5c00000386b) received by PID 14443 (TID 0x7f24b20f1500) from PID 14443; stack trace: ***
[...]

Issue #671, opened 2015-10-22
Support MPI synchronization in Vector


vincentchabannes:
keyword sync(...) allow to apply a synchronization of value of a vector.
Several kind of synchronization can be define in order to assign the dofs values shared by several proc : 
- operator = : value equal to one of them (generally the active process )
- operator + : value equal by summing all values for each proc which share the dof 
- operator min/max : value equal to the min/max value for each proc which share the dof

```
auto v = Vh->element();
// do some operations on v
sync( v, "=" );
sync( v, "+" );
sync( v, "min" );
sync( v, "max" );
```

If the vector v is modified only locally (e.g on a marker), the values synchronized
can be wrong value by deleting last insertion. For fix this, we can give
a list of dof index which are used to generate the new values of dof shared.

```
v.on(_range=markedfaces(mesh,"toto"),_expr=cst(3.1415));
std::set<size_type> dofUsed;
// fill dofUsed
sync(v,"=",dofUsed ); // also available for all operator
```

vincentchabannes:
it fix also the P1 visu of a P0d field in // (activated with --exporter.element-spaces "P0-P1" )

prudhomm:
:+1:

prudhomm:
It should now be closed no?

prudhomm:
Should it be closed @vincentchabannes? Or are there missing bits still?

PR #1416, opened 2020-01-23
Feature/toolboxes refactoring

vincentchabannes:

-----

Update in feelpp-lib : 
- operatorinterpolation : #1417
- feelvf : #1475
- feelvf : symbolic expression can depend on an expression which can depend on another expresssions etc..
- feelvf : support differentiation of expression with respect to a symbol name
- feelvf : refactoring of mat/vec keywords, replace boost::fusion by boost::hana.
- feelmodels : refactoring of ModelParameters (not only constant parameter)
- feells :  TO COMPLETE

Major update in feelpp-toolboxes : 
- core : new definition of toolbox thanks to ModelFields and ModelPhysics class. Factorize and simplify the definition of symbol and the genration of materials properties
- post-processing update : 
   -- export math expressions from JSON
   -- factorize code of post-processing in toolboxes
- material properties: 
     -- unique material properties share by mono/multiphysics toolboxes
      -- generate automatically all symbol expressions
      -- create a keyword that allows defining a material property for all materials (the expression is selected from the context, .i.e. the mesh element ) 
- heat toolbox : 
  -- new post-process measures : TO COMPLETE
- fluid toolbox :  
   -- no use a composite space for fluid/pressure
   -- implement rigid particle bc
- levelset/multifluid : 
  --  TO COMPLETE
- coefficient form pdes (new toolbox) : 
  -- a pde or a system of pdes can be solved by given the coefficients:  `m d^2 u/dt^2 + d du/dt +  \nabla \cdot (  -c \nabla u ) + \beta cdot \nabla u  + a u = f` where m, d, c, \beta, a, f are the coefficients.
  -- the type of finite element used in each equation can be differents


vincentchabannes:
@metivett test_levelset.cpp doesn't compile. Could you fix asap?

metivett:
> test_levelset.cpp doesn't compile. Could you fix asap?

@vincentchabannes Done ! I also merged some fixes in the `MultiFluid` toolbox from the `feature/ls` branch.
I am working on updating all the testcases (only `multifluid_square` is done at the moment).

vincentchabannes:
@metivett what is the goal of syncDofs (feelpp/feel/feeldiscr/syncdofs.hpp).
We have already the sync keyword which seems to do the same thing I think, no?

metivett:
> @metivett what is the goal of syncDofs (feelpp/feel/feeldiscr/syncdofs.hpp).
> We have already the sync keyword which seems to do the same thing I think, no? see #671 . And I have added also the `close` args with on(..) projection which can apply easily the sync.

Actually the goal was to allow a simpler interface (using only a lambda function instead of a functor derived from SyncOp), and to enable the synchronisation of a restricted set of dofs specified by an element range (and not a set of dofs). We could extend the `sync` function to provide these functionalities for sure, but I did not want to break other codes...

vincentchabannes:
I do not agree, the interface allow to support range, as you can see [here](https://github.com/feelpp/feelpp/blob/feature/toolboxes-refactoring/feelpp/feel/feeldiscr/functionspace.hpp#L4345). 
And the options with ComponentType::NO_COMPONENT and the boolean are optionally(optimization)

It's hard to maintain another mpi code.

metivett:
Ok, but the functor approach is a bit cumbersome I think, and the syncDofs function is more generic than the sync one since it allows to synchronise any template vector (the vector and datamap types are separate parameters, see eg [here](https://github.com/feelpp/feelpp/blob/feature/toolboxes-refactoring/feelpp/feel/feells/fastmarching_impl.hpp#L93) ).

vincentchabannes:
There is some optimisation for the classic case ( =) and I treat only the dofs on interprocess, this is not the case with your implementation (some loop over all range).
we can add easily an interface for your case.

metivett:
I don't really see what you can optimise further, at some point you need to find the interprocess dofs which are in your range (in my case it is done in the function while you do it separately with the functionspace->dofs() call). But if you want we can merge the two functions ;-) !

vincentchabannes:
1) I doesn't call localToGlobal (you do this  for all element)
2) The dofs at interprocess are identified only one time, we can reuse it after for each sync 
3) In the case of = operator,  we don't need to apply 2 passes of mpi communication, but only one.

metivett:
I don't really agree, I just think the purposes are just a bit different : 
1 -> if you don't call localToGlobal, then you need to send all the ghost interprocess dofs, not just the one which are in your range, which can lead to unneeded communications
2 -> the dofs which are both interprocess and in your range can't be reused (the range changes)
3 -> ok, but I had a generic lambda synchronisation in mind

But I will not fight to death for my function ;-), as I said we can refactor the sync one to enable my use (and maybe the flexibility of lambdas).

vincentchabannes:
For finish :)
1) No, the informations are in DataMap. but you need to communicate if the range is not contains the active dofs. Try with a random partitioning
2) Maybe in your case, but generally no.

I comment on this function because I need to review all mpi comm, there are some troubles with last boost (>=1.70) #1296  #1438.

metivett:
@vincentchabannes So I added the [`ModelMeasuresQuantity`](https://github.com/feelpp/feelpp/blob/feature/toolboxes-refactoring/toolboxes/feel/feelmodels/modelcore/modelmeasuresquantities.hpp) to deal with delayed evaluation of measured quantities. I did not wrap the tuples with another class like `ModelMeasuresQuantit**ies**`, as it is simpler than in the `ModelFields` case, but I can do it it necessary to uniformise the language. What do you think ?

vincentchabannes:
@metivett Yes, it sounds better (and probably necessary in some generic cases and reduce the tuple size). But there is no urgency.

metivett:
@vincentchabannes I also have a remark regarding the implementation of "delayed evaluation" in `ModelField`s: you did it by providing an `updateFunction`, in addition to the evaluation of the accessor (e.g. for "dirac" in `LevelSet`, we have `this->dirac( up=false)` and `bind( updateDirac, M_doUpdateDirac)` basically). However, I think this is a bit too complex to implement in the class (you need to check twice of the update condition, both in `dirac(bool)` and `updateDirac(bool)`, and for a non-specialised user, this can lead to misunderstanding.
I would propose to use the simpler approach I implemented in the `ModelMeasuresQuantities`, which consists in having one accessor (like e.g. `dirac(bool)`), which can update or not, and pass a function (like `bind( dirac, this, ref(doUpdateDirac) )`) to the `ModelField` directly (without the update function, which is already managed by `dirac(bool)`. The implementation of `ModelField` then just needs to be adapted for the functor case (if the template field is a field, you return it, and if it is a callable (I use `std::is_invocable`), you return the call). What do you think ? 
If that's ok for you, I can do the change !

metivett:
After careful look at the way the update is done at the moment, I also think this is not optimal in terms of efficiency, as it can lead to problems if you pass a field by value to `ModelField`. In this case, `ModelField::M_field` will be an `element_type` (not a ptr or ref), and if you have already updated the toolbox field (which means `M_doUpdate = false`), then the update function does nothing, so that `M_field` is not properly computed.
I really think we should only use one function, which possibly updates, and returns the value (which would be copied or moved in `ModelField::M_field` by `applyUpdateFunction`).

vincentchabannes:
@metivett I agree with a SymbolExpr struct instead of call the get<>, it's was in my mind but not find the time for this.
About the update function, I am not sure to understand the issue. I do this feature in being careful not to break anything (and it's only used by levelset). The mechanism of modelfields/symbolsexpr  is fully used, functional and not trivial. Also regarding the field passed by value, it completely not recommended, or even I would like to say it's forbidden.
But if you think that you can improve that, no problem but require to test a lot.

vincentchabannes:
@metivett and just for info if didn't see, there is no more GinacExVf, all symbolic expression are managed by GinacMatrix now.

metivett:
> @metivett I agree with a SymbolExpr struct instead of call the get<>, it's was in my mind but not find the time for this.
> About the update function, I am not sure to understand the issue. I do this feature in being careful not to break anything (and it's only used by levelset). The mechanism of modelfields/symbolsexpr is fully used, functional and not trivial. Also regarding the field passed by value, it completely not recommended, or even I would like to say it's forbidden.
> But if you think that you can improve that, no problem but require to test a lot.

Hi @vincentchabannes ! Ok, so as a first proposal, I did not change all the SymbolExpr field evaluation and update, but I changed the signature and approach of the (lazy) update of the fields in ModelField. The point here is to avoid breaking OOP by giving too large access into protected or private `updateField` toolbox functions. I reverted these functions to their original purpose, and only adapted the update mechanism of `ModelField`, which now overwrites its internal field (with something like `M_field = update();`, update being in fact a call to `toolbox->field( update=M_doUpdateField)`.
I pushed the changes into `feature/ls` (1b2d4bd3628ae972eec9149a0d3b733a58fed02f), so have a look when you can, and I will merge into `toolbox-refactoring` if you agree !

metivett:
Hi @vincentchabannes ! Regarding the `symbolsExpr` keyword (or prefix), I think maybe we should implement the standard replacement of all "-" to "_" in the keywords, as GiNaC incorrectly parses keywords of the form "bla-bla" as "bla" - "bla", and thus crashes. However, "bla-bla" is authorised as a model `prefix`, so for standard rule "-" -> "_" could help the user keep the same prefix and keyword while allowing for the use of symbols like "bla_bla_symbol". What do you think ?

vincentchabannes:
Hi @metivett . I agree but it requires changing a lot of cfg/json. It's why I added the keyword attribute. When this first part of the refactoring will be merged (there are few changes in our json/cfg),  I think we can try to change the maximum of things (I planned to review fully the BoundaryConditions) in order to have a stable format, version v1.0.  And then, a version control.

metivett:
Hi @vincentchabannes ! I think there might be an issue with the `ModelFields` update function mechanism, which I think is more precisely related to a `move` (or rvalue ref) somewhere. Indeed, the code crashes when using this update mechanism for a symbol used in some expression (eg in an `integrate` call), and according to the debugger, it seems to come from an uninitialised `shared_ptr M_field` (its `__ptr_` is null). However, that's not always the case (in my tests it happens for example only at the second use of the symbolExpr), so that I suspect something goes wrong regarding the copying or moving of the `modelField`. Obviously, using the `update( M_field )` you first implemented, which forwards to a toolbox update function with the guard 
`if( !field ) field.reset( new element );` hides the issue, but recreates the element at each update. Do you know where this could come from ? I am trying to detangle the copies and moves of the model fields' symbolsExpr, but it's not straightforward ;-)...

vincentchabannes:
@metivett I forgot the details in my implementation but indeed there are small cautions with rvalue, maybe a bug. Are you sure that the shared pointer is stored in your class, else it's only in ModelFields and generally it's a temporary object. Can you try to write a small code that reproduces the issue?

vincentchabannes:
@metivett I have merged feature/ls but now there is a compilation error, see https://buildkite.com/feelpp/feelpp/builds/3494#f88fa3a0-ce43-4de5-b569-63f789c0d9a0/194-1738  : 
```
/home/feelpp/src/feelpp/toolboxes/feel/feelmodels/levelset/levelset.cpp:667:12: error: function 'leftCauchyGreenTensorExpr' with deduced return type cannot be used before it is defined
```

vincentchabannes:
@metivett Can you try to fix the compilation error? I need to check the full CI.

metivett:
> @metivett Can you try to fix the compilation error? I need to check the full CI.

Sure, sorry for the delay; I'm on it !

vincentchabannes:
@metivett there are compilation errors in multifluid

```
feelpp/toolboxes/feel/feelmodels/multifluid/multifluid.hpp:214:51: error: no matching function for call to 'modelField'
                                                  modelField<FieldCtx::ID>( FieldTag::globalLevelsetElt(this), prefixvm( prefix, "global-levelset"), "phi", this->globalLevelsetElt(false), "phi", this->keyword(),
```

Issue #1485, opened 2020-06-29
Minmax returns erroneous min value when run on a range which does not intersect all processes


metivett:
When computing the `minmax` of an expression on a custom range, the returned `min` can be wrong (and equals 1e-30) if the range does not intersect one or more partitions.

PR #1486, opened 2020-07-03
Fix/minmax

metivett:

-----

Fixes a bug with `minmax` which can return a wrong `min` value when run in parallel on a range which does not intersect all partitions (issue #1485).


feelpp/feel/feelvf/evaluator.hpp
```diff
@@ -891,7 +891,7 @@ BOOST_PARAMETER_FUNCTION(
 
     int indexmin = 0;
     int indexmax = 0;
-    double mine = 1e-30;
+    double mine = 1e+30;
```
vincentchabannes, line 894: it will better to use `std::numeric_limits<double>::max()`

feelpp/feel/feelvf/evaluator.hpp
```diff
@@ -891,7 +891,7 @@ BOOST_PARAMETER_FUNCTION(
...
     double maxe = 1e-30;
```
vincentchabannes, line 895: and here  `std::numeric_limits<double>::epsilon()`

metivett: @vincentchabannes @prudhomm Actually, I think `std::numeric_limits<double>::lowest()` would even be better. Otherwise, I think the `max` might also be wrong if we want to compute the max of a negative field on multiple partitions, no ?

vincentchabannes: Indeed, I took a look quickly this morning and I was quite surprised by the use of absolute value for the max (which explains the use of lowest positive value but I didn't take the time to understand). I think you are more inside that me to understand this issue

PR #1523, opened 2020-09-14
Patch: fix eigen lapack linkage

metivett:
This short patch fixes `feelpp` compilation, which can fail when linking `libeigen_lapack_static` from `contrib/eigen`. This comes from a missing `target_link_libraries`.


prudhomm (approved):
thanks!

Issue #1526, opened 2020-10-04
Provide interface for partitioning locally connected faces


prudhomm:
For some PDEs we have to deal with constant functions (scalar or vectorial). 
They have usually local support associated to one or several physical face markers for example in integral boundary conditions or when modeling rigid body motion.
In order to accomodate the fact that these functions usually couple the degrees of freedom associated to some other discretisation, it may be beneficial to have the elements associated to the faces where the constant functions are defined to be neighbors of each others in order to avoid MPI communication and improve scalability

> update `feelpp_mesh_partitioner` documentation

here is a command line
```
 ./feelpp_mesh_partitioner --json cases/three-spheres-part.json --ifile cases/three-spheres.geo  --part 24
 ```

 and to export to paraview
 ```
 mpirun -np 24 ./feelpp_mesh_exporter --gmsh.filename three-spheres_p24.json
 ```

**Comment by romainhild** [comment]
@prudhomm is there an example somewhere as to how to use the partitioner now ?

**Comment by prudhomm** [comment]
here is a command line
```
 ./feelpp_mesh_partitioner --json cases/three-spheres-part.json --ifile cases/three-spheres.geo  --part 24
 ```

 and to export to paraview
 ```
 mpirun -np 24 ./feelpp_mesh_exporter --gmsh.filename three-spheres_p24.json
 ```

prudhomm:
@romainhild the testcase is in `feelpp/tools/mesh/cases`

Trophime:
@prudhomm would it be possible to export process ids in feelpp_mesh_exporter to view partition within paraview?

Trophime:
@prudhomm what is the meaning of the "size" parameter in the json defining the partioner aggregation?

thomas-saigre:
> @prudhomm would it be possible to export process ids in feelpp_mesh_exporter to view partition within paraview?

@Trophime I did this modification in https://github.com/feelpp/feelpp/commit/2f01fb2f86c9b0c9ecfd6870f3f01bfe903bea0d (#2222) : the pid is exported when the application `feelpp_mesh_exporter` is used

Trophime:
@thomas-saigre  Cool. Do you have an example? I need to add an "export_pid" right?

thomas-saigre:
`mpirun -np 4 feelpp_mesh_exporter --gmsh.filename <path/to/your/mesh> --dim <dim>` should work. But this is not merged yet, so you will need to recompile from the branch.
[...]

PR #1537, opened 2020-11-10
Remove soption("json") before checking if present

romainhild:
@prudhomm I guess you forgot to remove this line, the app crashes if the option is not used

### DIFF
--- a/feelpp/tools/mesh/mesh_partitioner.cpp
+++ b/feelpp/tools/mesh/mesh_partitioner.cpp
@@ -90,7 +90,6 @@ int main( int argc, char** argv )
         return 0;
     }
     nl::json partconfig;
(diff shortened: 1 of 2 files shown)


romainhild:
this is linked with #1526, but even with this change, we cannot use the app as before anymore.

feelpp/tools/mesh/mesh_partitioner.cpp
```diff
@@ -90,7 +90,6 @@ int main( int argc, char** argv )
         return 0;
     }
     nl::json partconfig;
-    Feel::cout << "reading partitioner configuration: " << ( fs::current_path() / soption("json") ) << std::endl;
```
prudhomm, line 93: indeed

prudhomm (approved):

PR #1622, opened 2021-05-28
Feature/meshstructured

vincentchabannes:

-----

MeshStructured features :
- 1d/2d/3d with hypercube element (TODO : Simplex)
- geometry currently available: `hyperrectangle`. The setup of the geometry is  given by a json file, see  examples in test_meshstructured. (TODO : other shapes as for example an arbitrary Quadrangle(2d) or QuadrilaterallyFacedHexahedron(3d) by using geo transformation of hypercube).
- partitioning currently available: `striped`  one direction (TODO : by block)

@prudhomm It will break some code that uses MeshStructured : 
- template parameters of class MeshStructured
- constructor args

### DIFF
--- a/testsuite/feeldiscr/test_meshstructured.cpp
+++ b/testsuite/feeldiscr/test_meshstructured.cpp
@@ -25,279 +25,247 @@
    \date 2015-04-27
  */
 
-#define USE_BOOST_TEST 1
-#if defined(USE_BOOST_TEST)
 #define BOOST_TEST_MODULE test_meshstructured
 #include <feel/feelcore/testsuite.hpp>
-#endif
 
-#include <feel/feelfilters/loadmesh.hpp>
+#include <feel/feeldiscr/meshstructured.hpp>
+//#include <feel/feelfilters/loadmesh.hpp>
 #include <feel/feeldiscr/pch.hpp>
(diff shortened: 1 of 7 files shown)


testsuite/feeldiscr/test_meshstructured.cpp
```diff
@@ -25,279 +25,247 @@
...
+
+FEELPP_ENVIRONMENT_NO_OPTIONS
 
-namespace Feel
-{
 
-inline
-AboutData
-makeAbout()
+#if 0
```
prudhomm, line 56: the test does not work with coordinates ?

prudhomm (approved):
Is there a test for the case with coordinates ? I found one but it is commented out

prudhomm:
nice job for refactoring this

testsuite/feeldiscr/test_meshstructured.cpp
vincentchabannes: the coordinate test is [here](https://github.com/feelpp/feelpp/blob/feature/meshstructured/testsuite/feeldiscr/test_meshstructured.cpp#L213-L269). This comment is a hbf test but it was always commented before my change.

PR #1671, opened 2021-07-19
fix geometry of thermal-fin

thomas-saigre:
add fins in layers and fix template
[ci skip]

-----


romainhild (approved):

prudhomm:
is the application installed with the documentation ?
why in mor/toolbox-mor? 
I would suggest 
- toolboxes/generators/cases/thermal-fin -> geometry + cfg/json
no ?

also the generated geo file should contain

/cc @thomas-saigre @romainhild

thomas-saigre:
@prudhomm I documented in https://github.com/feelpp/feelpp/blob/saigre/fix-thermal-fin/mor/toolbox-mor/thermal-fin/case_generator.adoc
I can move those files in the drectory `toolboxes/generators/cases/thermal-fin` if it make more sense than where it is now. I think I put it in `mor` becausethe case was build for this toolbox, but we can indeed use is with another toolbox

prudhomm:
you generate the json file as well ?

prudhomm:
we could be indeed interested in different physics, eg solid and fluid mechanics

thomas-saigre:
> 
> 
> you generate the json file as well ?

Yes the json is part of the templates

prudhomm:
the application could generate the json for different physics, in case of fluid mechanics the fin needs to be in a box

prudhomm:
`feelpp_case_generator` should be renamed `feelpp_case_generator_thermal_fin`

PR #1691, opened 2021-09-22
Feature/toolboxes refactoring hdg

romainhild:

-----

Refactoring of the MixedPoisson Toolbox to be as close as possible to the other toolboxes.
Modification of coupled mixed poisson, poroelastic and thermoelectric to work with the latest version of the toolbox.
Add tests for each application.


feelpp/feel/feelmodels/modelmarkers.cpp
```diff
@@ -65,5 +65,14 @@ ModelMarkers::setPTree( pt::ptree const& p, ModelIndexes const& indexes )
...
 }
 
+bool
+ModelMarkers::empty() const
+{
+    if( this->size() > 1 )
+        return false;
+    if( this->size() == 0 )
+        return true;
+    return this->begin()->empty();
```
vincentchabannes, line 75: why overload this function (it's already in the parent class) ?

feelpp/feel/feelmesh/convex.hpp
```diff
@@ -106,5 +106,10 @@ class Convex : public ConvexBase
 private:
 
 };
+
+template<typename Convex>
+struct ReduceDim {
```
vincentchabannes, line 111: I think, it should be move inside simplex/hypercurbe class, and maybe add a static check about a null dim.

But do you know that you have the type `trace_mesh_type` in Mesh. I think it's exactly what you want,no? (mesh on topological face) If yes, you can remove these additions.
Moreover, you can write `Mesh::trace_mesh_type::trace_mesh_type` in 3d for a mesh defined on edges

feelpp/feel/feelmodels/modelmarkers.cpp
romainhild: if you have one marker equals to the empty string `""`, then the set is not empty but the markers should be empty.

feelpp/feel/feelmesh/convex.hpp
romainhild: Indeed, I didn't know about `trace_mesh_type`, I'll revert these changes and use that. Thanks

PR #1739, opened 2021-12-07
Add basic finite element matrices construction in python

### LINKED ISSUE #1738: Add basic finite element matrices construction in python
in python we should be able to build basic finite element matrices like mass and stiffness matrices simply 

/cc @thomas-saigre

prudhomm:
Fixes #1738

other matrices can be added in the future

skip components

-----


prudhomm:
@vincentchabannes the PR is only on the python version, no need to check the toolboxes or mor
some tests/examples have been added

pyfeelpp-all/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.3)
+cmake_minimum_required(VERSION 3.21)
```
vincentchabannes, line 1: this is a big constraint, most OS doesn't are to this level.

prudhomm: this is not used currently and will be changed

prudhomm: this is used only by cmake preset which works only at cmake 3.21.
python codes will move there and the cmakelists,txt will be updated properly

prudhomm: it was added by mistake, I can of course lower the version but it has not impact for now except for those using cmake --preset

vincentchabannes: So OK, If you are sure that cmake doesn't failed for a basic use (without preset and thus also lower version)

prudhomm: the main cmakelist does not `add_subdirectory` pyfeelpp-all for now (only via preset)

vincentchabannes (approved):

Issue #1740, opened 2021-12-07
move feel/feelcrb in component mor


prudhomm:
in the effort of #1734, we shall move feel/feelcrb to mor

- [ ] move directory
- [ ] check cmake files
- [ ] move tests depending on feel/feelcrb

/cc @romainhild @vincentchabannes @thomas-saigre

PR #1741, opened 2021-12-07
Prudhomm/issue1740

prudhomm:

-----

implements #1740 

this is the reorganisation of the component. Files has been moved around from feelpp 
and testsuite components regarding crb and mor component has been (hopefully better) reorganized.
the PR does not involve any change in implementation


prudhomm:
@thomas-saigre in pyrb you are only modifying the python layer right ? or are you modifying also the C++ layer (not the wrapper I mean) ?

thomas-saigre:
> @thomas-saigre in pyrb you are only modifying the python layer right ? or are you modifying also the C++ layer (not the wrapper I mean) ?

yes, instead if one functionality has not yet been implemented (as it was the case with the computation of mass matrix for more). Right now I don't have any problem, but I'll let you know if it is the case

prudhomm:
the compilation failed in mor 

https://buildkite.com/feelpp/feelpp/builds/4080#c98a7814-2b92-4a85-8d0d-57faee383674/263-453

locally it compiles, I will check asap the issue

romainhild:
@prudhomm all the history of the files have been deleted !
We need to keep the history and tell git that the files have just moved.

Regarding the error of compilation, you moved all the hpp in mor/mor/feelmor but did not change the includes in the apps. I don't know how it could compile locally.

prudhomm:
Ok Indeed that's not good. I thought got was good at detecting renames/moves . apparently it is not the case.

As to compilation it used the version installed

PR #1801, opened 2022-03-09
replace boost::filesystem by std::filesystem

metivett:
Replaces boost::filesystem with std::filesystem to reduce boost dependency.

-----


prudhomm:
merged with develop
I will review this weekend

prudhomm:
@metivett this fails https://buildkite.com/feelpp/feelpp/builds/4323#283f4c92-8da0-4c22-9f1f-24a64793afbc/6-11239
I will check the dependencies asap
there is no one-to-one correspondence between boost::filesystem and std::filesystem so it may not be just switching boost:: to std::

metivett:
> @metivett this fails https://buildkite.com/feelpp/feelpp/builds/4323#283f4c92-8da0-4c22-9f1f-24a64793afbc/6-11239 I will check the dependencies asap there is no one-to-one correspondence between boost::filesystem and std::filesystem so it may not be just switching boost:: to std::

Indeed, that's why I had to change a few things in https://github.com/feelpp/feelpp/pull/1801/commits/caeed9df04439b1c55df463a9d957d49d53f8839.
However, I don't understand why we have the `error: no namespace named 'filesystem' in namespace 'std'` in buildkite, there should be such a namespace...

prudhomm:
@metivett it is probably a problem with the GCC version. We use clang but clang uses libstdc++ and I think that std::filesystem is available only in GCC9. 
I have to check the dependency version on GCC

prudhomm:
@metivett I just checked and we use GCC 9. Also `std::filesystem` is properly defined for c++17 in GCC9
see eg ` /usr/include/c++/9/bits/fs_fwd.h`
I will have a closer look, perhaps later today

prudhomm:
@metivett in the mean time you can start docker 
```
docker run --rm -it ghcr.io/feelpp/feelpp-env:ubuntu-20.04
```
get your branch and compile to see what is going on

vincentchabannes:
@prudhomm this PR is ready.

prudhomm (approved):

Issue #1489, opened 2020-07-06
Enrich range object


vincentchabannes:
A range represent a set of entity that we can iterate.
Adding various information sinside the range structure can be very useful. Even if the range is empty, we can have some informations. For example the mesh and optionally a mesh support. 

Some impacts : 
* allow to get by default the good WorldComm when we do integrate(_range=...,_expr=..).evaluate()
* integrate over internal surfaces but defined at the boundary of mesh support.

PR #1830, opened 2022-04-13
Feature/ensemble

### LINKED ISSUE #1825: support ensemble simulations in parallel
support ensemble simulations in parallel to accelerate the evaluations of models in eg kalman filters

if we have say N*P processes, we can split the world communicator into N sub communicators grouping P processes that can execute in parallel. It is expected that toolboxes can. be run in this context. It means that they should support subcommunicators and that we can disable printing to avoid scrambling the screen or logs. We could just enable printing for only one of the subcommunicator

### LINKED ISSUE #1826: split worldcomm in n worldcomms
splitting a worldcomm into n subworldcomm having each say p processes enables parallel methods for ensemble simulations, reinforcement learning or parameter sweeps

### LINKED ISSUE #1827: support WorldComm in norm computations other than Environment::WorldComm
norm computations should be computable on any relevant communicator not only Environment::WorldComm 
useful for #1825

### LINKED ISSUE #1828: support ini parser in python
a ini parser is useful to manipulate easily options

@Trophime I used config parser and I add a section `feelpp` for section-less options

### LINKED ISSUE #1829: support toolbox worldcomm in statistics/measures computations
measures and statistics use Environment:worldCommPtr() instead of the toolbox. It then forbids #1825 or have a specific communicator for a toolbox

prudhomm:

-----
- fixes #1825
- fixes #1826 
- fixes #1827 
- fixes #1828
- fixes #1829


prudhomm:
@lberti @VLCeline @zakaryael we should now be able to do RL in parallel

@lberti idem for ensemble simulations

prudhomm:
@vincentchabannes I added some code to tests the toolbox in // in `toolboxes/tests/heat/test_heat.cpp`

vincentchabannes:
@prudhomm I think really that is not a good way to fix the communicator issue. And there are other locations where the comm will be wrong. If we implement #1489, changes will be only in feelpp, not in toolboxes, and fix all parts. Take more time but impact clearly larger

prudhomm:
Ok i will try to do that

vincentchabannes:
Moreover, I have planned to not apply mpi reduce when computing each measure, but only one mpi comm when all local measures will be stored

vincentchabannes:
@prudhomm  for the struct with range, the idea is to store also the MeshBase object inside (even if the range is empty on a process). So we will access some useful info as WorldComm, markers... .

PR #1787, opened 2022-02-25
Feature/pyrb

thomas-saigre:
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/feelpp/feelpp/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you written new tests for your changes?
- [ ] Have you successfully run the Feel++ testsuite with your changes locally?
- [x] Have you written Doxygen comments in your contribution ?

-----

- closes #1732
- closes #1675
- implement reduced basis method in python using feelpp and petsc4py
- closes #1768
- closes #1888
- #1883

This PR contains the module reducedbasis use the run the offline and online part for the reduced basis method.
For now, only steady-time cases can be treated, and only "coercive" problem, as long as the scm module has not yet been wrapped in python, so it uses the min-theta approach for the computation of error bound.
This is not pushed yet, but as soon I have tested it will be, and different outputs can be studied (for the not-compliant case).

The content has two main parts :

The python module reducedbasis contains all the classes and code for offline and online parts
The script generate_basis.py allows to generate a reduced basis using one of 3 algorithms (from a fixed random sampling, from a POD construction, of the greedy algorithm)
There are still some issues in the tests pipeline due to persistent files not compatible between 2d and 3d cases #1768, which I did not manage to resolve, but when I run by hand all the tests locally, deleting by hand the directory feel/crbdb and `feel/pyfeelppmor-tests, the tests pass well both in sequential and parallel.

PR #1840, opened 2022-04-21
Add outputs for ToolboxMor and use CRBModels

### LINKED ISSUE #1797: Use a model for CRB applications
As for the toolboxes, we should be able to define a CRB applications from a json file.
The informations could be:
- the parameters:
  - name of the parameter
  - min/max of each parameter
  - possibly the sampling
- the outputs:
  - name of the output
  - type of output
  - necessary informations to build the linear form associated (dimension, field, markers, coordinates,...)

Those informations could be used in ToolboxMor, which should also be renamed, to build the parameter space and the outputs.
The informations to build the affine decomposition should stay in ModelProperties if needed.

romainhild:

-----

- Fixes #1797
  - add class CRBModelProperties, CRBModelParameters and CRBModelOutputs
  - ParameterSpace can be created from a CRBModelParameters
  - In the json, we need to add a section `CRBParameters` which contains parameters with min and max values
- Add SensorPointwise to evaluate a field at a certain point
- ToolboxMor uses CRBModelOutputs to create outputs from a section `CRBOutputs` in the json
  - Types of outputs currently available: mean, integrate, sensor, point
  - mean and integrate take `markers`,  `topodim`, and optionaly `expr`
  - sensor takes `coord` and `radius`
  - point takes `coord`
  - see examples [here](https://github.com/feelpp/feelpp/blob/cacb236c6f5b1529b5b08f209c7f5fd0e65bd37e/mor/apps/opusheat/opusheat-heat.json#L143-L169)
- ToolboxMor uses DEIM for the outputs of type mean and integrate if the `expr` is dependant on the parameters.
  - the `expr` should use `crb_u`, `crb_grad_u_{0,1,2}` or `crb_dn_u` and the CRBParameters for the symbols
  - DEIM now creates its own options depending on the prefix at the construction
- Add convergence tests for `feelpp_mor_heat` app on the field and outputs

For now, the outputs works only when using a scalar function space.


romainhild:
@thomas-saigre some changes here will impact you, you should check if everything is working for you.

romainhild:
@prudhomm @vincentchabannes can you review the PR ? 
@thomas-saigre you should look at the changes in this branch, I have modified some of your tests

thomas-saigre:
> @thomas-saigre you should look at the changes in this branch, I have modified some of your tests

so if I understand well, we define parameters in the JSON by two ways :
1. `CRBParameters`, telling the range of values
2. `Parameters` telling the "default value"

Maybe I will have to modify some code in what I've done in [feature/pyrb on fb7104a57a20e4b842367b193d2b5507afac5650 ](https://github.com/feelpp/feelpp/pull/1787/commits/fb7104a57a20e4b842367b193d2b5507afac5650#diff-479a0b26693b32a19f8e20ab02b265006e7f5578bd9c7ba09456bccf2d2437afR85-R86)

romainhild:
Probably yes, but there is a value field in CRBModelParameter, so you could use this for the default value. But there is no `toParameterValues()` function in CRBModelParameters, since the value is not mandatory.
Anyway, you shouldn't use the toolbox to create a ModelParameters.

thomas-saigre:
as long as there is an easy way to get the default values, I'm fine with it !

romainhild:
you can do 
```python
for name, param in parameters:
    print(param.value())
```
But you should also modify the json so that the value is in the CRBParameters section.

vincentchabannes (approved):

prudhomm (approved):

PR #1889, opened 2022-06-30
Feature/optimization symbolic expr degree

vincentchabannes:

-----

* Optimization of the computation total degree of a symbolic expr (polynomial). Previously the algo complexity was clearly very bad ( double loop + recursion non-terminal). Now it's just an iterative version which is faster. For example, this computation was done with EIM->beta() online computation and the perf is huge.
* Reduce some log verbosity


feelpp/feel/feelvf/detail/ginacmatrix.hpp
```diff
@@ -1125,7 +1153,7 @@ private :
         if ( M_isNumericExpression )
             return;
 
-        std::vector<std::pair<GiNaC::symbol,int>> symbTotalDegree;
+        std::vector<std::pair<GiNaC::symbol,uint16_type>> symbTotalDegree;
```
prudhomm, line 1156: warning could you keep int ? 
I am making in another branch the polynomial degree dynamic like in Eigen (compile time or dynamic)

prudhomm: here is a discussion also on uint vs int http://eigen.tuxfamily.org/index.php?title=FAQ#Why_Eigen.27s_API_is_using_signed_integers_for_sizes.2C_indices.2C_etc..3F

prudhomm (approved):

feelpp/feel/feelvf/detail/ginacmatrix.hpp
vincentchabannes: I am not sure to understand why it's a problem because, here, it's a local variable used in the algorithm (and I need to check if value is invalid_v< uint16_type> or not. If really an issue, we can change to int and check non negative value.

PR #1890, opened 2022-06-30
 Support Eigen3 external dependency

vincentchabannes:

-----

the cmake option FEELPP_USE_EXTERNAL_EIGEN3=ON  allow to use an eigen3 as an external dependency.
@prudhomm enable this option  fix tests on debian11 because Gmsh is compiled with Eigen3 on system and there is a conflict)

### DIFF
--- a/feelpp/contrib/CMakeLists.txt
+++ b/feelpp/contrib/CMakeLists.txt
@@ -169,6 +169,13 @@ endif()
 
 
 include(FetchContent)
+
+option( FEELPP_USE_EXTERNAL_EIGEN3 "USE_EXTERNAL_EIGEN3" OFF )
+if ( FEELPP_USE_EXTERNAL_EIGEN3 )
+  FIND_PACKAGE(Eigen3 REQUIRED)
+  add_library(Eigen3::Eigen ALIAS eigen)
+else()
+
 set(EIGEN_BUILD_DOC OFF)
 set(BUILD_TESTING OFF)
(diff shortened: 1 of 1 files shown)


feelpp/contrib/CMakeLists.txt
```diff
@@ -693,7 +701,7 @@ if ( FEELPP_ENABLE_TASKFLOW )
   message(STATUS "[feelpp-taskflow] cmake dir: ${CMAKE_INSTALL_DATADIR}/feel/cmake/modules")
   FetchContent_MakeAvailable(taskflow)
   add_library(feelpp_taskflow INTERFACE)
-  set_target_properties(Taskflow PROPERTIES OUTPUT_NAME "feelpp_taskflow")
+  #set_target_properties(Taskflow PROPERTIES OUTPUT_NAME "feelpp_taskflow")
```
prudhomm, line 704: why is it related to the eigen3 dependency issue ?

prudhomm: I usually rename the libraries to avoid conflicts on the filesystem if there is also an installation of the library

prudhomm (approved):

feelpp/contrib/CMakeLists.txt
vincentchabannes: it's not related. I worked on debian11 and cmake give an error because target Taskflow is an interface lib (and so no OUTPUT_NAME)

Issue #1961, opened 2022-09-30
Add mesh quality as a trigger to remeshing in toolboxes


lberti:
In toolboxes, currently, remeshing is performed at a fixed frequency or at certain events.
Checking the minimum of the mesh quality helps to remesh only when necessary.
- [x] modify modelmeshadaptation.cpp with the new json specification
- [x] modify modelmesh.hpp with the condition triggering remesh

PR #1966, opened 2022-10-04
remeshing triggered by minimum mesh quality #1961

lberti:

-----

### DIFF
--- a/toolboxes/feel/feelmodels/modelcore/modelmeshes.hpp
+++ b/toolboxes/feel/feelmodels/modelcore/modelmeshes.hpp
@@ -183,7 +183,7 @@ private:
  * @tparam IndexType
  */
 template <typename IndexType>
-class ModelMesh
+class ModelMesh: public std::enable_shared_from_this<ModelMesh<IndexType>>
 {
 public :
     using index_type = IndexType;
@@ -320,6 +320,8 @@ public :
 
             std::set<std::string> const& requiredMarkers() const { return M_requiredMarkers; }
 
+            void setCurrentQualityValue(double q_min){M_min_quality_mas = q_min;}
+
             template <typename SymbolsExprType = symbols_expression_empty_t>
             bool isExecutedWhen( std::shared_ptr<Event> e, SymbolsExprType const& se = symbols_expression_empty_t{} ) const
                 {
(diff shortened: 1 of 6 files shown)


toolboxes/feel/feelmodels/modelcore/modelmeshes.hpp
```diff
@@ -360,15 +373,18 @@ public :
             static tabulate_informations_ptr_t tabulateInformations( nl::json const& jsonInfo, TabulateInformationProperties const& tabInfoProp );
 
             friend struct Execute;
+            
+            double M_min_quality_mas=0.9;
```
prudhomm, line 377: @lberti what does `mas` mean? shouldn't it be threshold ?

toolboxes/feel/feelmodels/modelmesh/meshale.hpp
```diff
@@ -255,6 +259,8 @@ public :
 
     void applyRemesh( mesh_ptrtype const& newMesh, std::vector<std::tuple<std::string,range_elements_type>> const& rangeElt );
 
+    void attachModelMeshesptr(std::shared_ptr<ModelMesh<index_type>> const& modelMesh_ptr);
```
prudhomm, line 262: rename to `attachModelMeshes` ?

toolboxes/fluid/cases/TurekHron/moving_body/cfd3.json
```diff
@@ -74,6 +74,7 @@
                     "events":{
                         "each_time_step":{
                             "frequency":10
+                            // "min_quality":0.4
```
prudhomm, line 77: why is it commented out ? would it be possible to add a test for this ?

prudhomm (approved):
@lberti it seems ok to me but there are a few update to do I think

vincentchabannes (changes requested):
@lberti this is a first step but still a lot of thing to handle.
 We probably need to change the weak ptr with a raw ptr in order to avoid the shared_from_this inheritance. The observer desgin pattern should be improved, and the event mesh_quality should be handle when the mesh is moved (and even before), not in the event each_time_step. Also, I think here the mesh quality is evaluated at each time, it shouldn't , only when the json say that this event can appear. The json section should be also improved.

prudhomm:
@vincentchabannes what do you mean by should be improved? what is wrong with the current implementation of the observer pattern ? may be I should have a look more closely and not just the diff.
regarding the computational efficiency I agree.

PR #2124, opened 2023-06-26
Resolves: enable dim != realdim for bvh

### LINKED ISSUE #2123: Enable dim != realdim for BVH
Enable the construction of Bounding Volume Hierarchies from surface-only mesh data

- [x] #2152

lberti:
Enable BVH to be constructed for surface-only meshes in 3d
- closes #2123

### DIFF
--- a/feelpp/feel/feelmesh/bvh.hpp
+++ b/feelpp/feel/feelmesh/bvh.hpp
@@ -41,14 +41,20 @@ namespace Feel
         return (n * epsil) / (1 - n * epsil);
     }
 
-    template<int nDim>
+    template<int nDim, int realDim>
     class BVHTree
     {
-        using self_type = BVHTree<nDim>;
-        typedef Simplex<nDim,1> convex_type;
-        typedef Mesh<convex_type> mesh_type;        
-        typedef typename mesh_type::trace_mesh_type trace_mesh_type;
-        typedef typename mesh_type::trace_mesh_ptrtype trace_mesh_ptrtype;
+        using self_type = BVHTree<nDim,realDim>;
+        using convex_type = Simplex<nDim,1,realDim>;
+        using mesh_type = Mesh<convex_type>;     
+        using trace_mesh_type = typename  std::conditional< nDim==realDim,
+                                                    typename mesh_type::trace_mesh_type,
+                                                    mesh_type >::type;
+        using trace_mesh_ptrtype =  typename  std::conditional< nDim==realDim,
+                                                        typename mesh_type::trace_mesh_ptrtype,
+                                                        typename mesh_type::ptrtype >::type ;
+        // typedef typename mesh_type::trace_mesh_type trace_mesh_type;
+        // typedef typename mesh_type::trace_mesh_ptrtype trace_mesh_ptrtype;
         typedef typename matrix_node<double>::type matrix_node_type;
 
         // Information on the primitive (bounding box, index, centroid)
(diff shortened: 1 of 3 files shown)


feelpp/feel/feelmesh/bvh.hpp
```diff
@@ -314,12 +320,12 @@ namespace Feel
 
             std::pair<bool,double> checkLeafIntersection(BVHRay const& rayon, trace_mesh_ptrtype mesh, std::vector<BVHPrimitiveInfo>& primitiveInfo)
             {             
-                if( nDim ==2)
+                if( realDim ==2)
```
vincentchabannes, line 323: use `if constexpr (..)`

feelpp/tools/mesh/mesh_partitioner.cpp
```diff
@@ -47,7 +47,8 @@ int main( int argc, char** argv )
 	po::options_description meshpartoptions( "Mesh Partitioner options" );
 	meshpartoptions.add_options()
         ( "dim", po::value<int>()->default_value( 3 ), "mesh dimension" )
-        ( "shape", po::value<std::string>()->default_value( "simplex" ), "mesh dimension" )
+        ( "topdim-equal-realdim", po::value<bool>()->default_value( true ), "mesh dimension" )
```
vincentchabannes, line 50: I think this is not the good way. @lberti could you replace it with an  option `realdim` (integer with no default value). And if not given, put realdim=dim and else just check that nRealDim >= nDim

feelpp/feel/feelmesh/bvh.hpp
```diff
@@ -314,12 +320,12 @@ namespace Feel
...
                 {
                     auto nodes = mesh->element( primitiveInfo[this->firstPrimOffset].M_primitiveNumber ).vertices();                                        
                     return checkIntersectionWithSegment(nodes,rayon);
                 }                
-                else //if ( nDim==3)
+                else //if ( realDim==3)
```
vincentchabannes, line 328: idem

feelpp/feel/feelmesh/bvh.hpp
```diff
@@ -343,7 +349,7 @@ namespace Feel
                                 std::vector<int> &orderedPrims)
         {
             LOG(INFO) <<fmt::format("cut dimension {}, start index primitive {}, end index primitive {}",cut_dimension,start_index_primitive,end_index_primitive);
-            Eigen::VectorXd M_bound_min_node(nDim),M_bound_max_node(nDim);
+            Eigen::VectorXd M_bound_min_node(realDim),M_bound_max_node(realDim);
```
vincentchabannes, line 352: use static vector eigen instead of dynamic . This is much more efficient. And if possible, put in cache (i.e. only one allocation during the alog)

And remove the `M_` prefix, it's reserved for attribute member, not local variable.

feelpp/feel/feelmesh/bvh.hpp
```diff
@@ -373,9 +379,9 @@ namespace Feel
             }
             else{
                 // Create a node, since there are at least two primitives in the list
-                node->buildInternalNode(current_parent,(cut_dimension+1)%nDim,
-                                    recursiveBuild( node, (cut_dimension+1)%nDim, start_index_primitive, mid, orderedPrims),
-                                    recursiveBuild( node, (cut_dimension+1)%nDim, mid, end_index_primitive, orderedPrims));
+                node->buildInternalNode(current_parent,(cut_dimension+1)%realDim,
```
vincentchabannes, line 382: an optimisation will be to precompute only one time this operation (some operation as division, modulo can be a little expansive). Maybe can be also done at compilation time, no?

vincentchabannes (changes requested):
I also noticed that the line `BVHNode * node = new BVHTree::BVHNode();` allocate dynamically several objects but I don't find any destructor of these object. This is done elsewhere? If no, there is a memory leak.

PR #2218, opened 2024-02-22
add quadrature order in sensor gaussian class

thomas-saigre:
/cc @prudhomm

-----

### DIFF
--- a/feelpp/feel/feeldiscr/sensors.hpp
+++ b/feelpp/feel/feeldiscr/sensors.hpp
@@ -249,9 +249,10 @@ public:
     {
         auto v = this->space()->element();
         auto phi = this->phiExpr( mpl::int_< space_type::nDim >() );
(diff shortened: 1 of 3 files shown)


thomas-saigre:
Here are some results of what has changed thanks to these modifications. I'm solving the equation $-\Delta u = f$ where a Gaussian sensor gives $f$, with Dirichlet boundary conditions I'm solving this equation for various orders of discretization (from 1 to 6).
Before the addition of quadrature order inside the computation of the sensor, we noticed that the quadrature error was too great and therefore lost the amelioration due to higher order of discretization:
Capture d’écran du 2024-02-29 09-21-16
On the other hand, with a higher order of quadrature, we see improvement.
Capture d’écran du 2024-02-29 09-22-45
**Note:** we reach the machine error for $p\geqslant 4$

prudhomm (approved):

PR #2359, opened 2025-06-17
2346 fix extractzipfile function in zipcpp

### LINKED ISSUE #2346: Fix extractZipFile function in zip.cpp
The `extractZipFile()` function does not check that the extraction path actually exists.

Source file: `feelpp/feel/feelcore/zip.cpp`.

We need to add this function call: `fs::create_directories(extractionPath.parent_path());`

- It creates all parent directories required for the given path.
- If the folder already exists, it does nothing and returns no error.
- If one of the parent folders doesn't exist, it creates it.

Philand:

- closes #2346 
-----
## Work done

- Improved the extractZipFile function. The function wasn't perfectly functional, as it didn't create the necessary path to the extraction file when the extraction file's parent folders didn't exist
- I added two tests to test this unzip function. One that simply tests unzipping and one that tests unzipping with download of a girder dataset.
- I run the test locally and it worked.


vincentchabannes (approved):

feelpp/feel/feelcore/zip.hpp
```diff
@@ -1,5 +1,11 @@
 #pragma once
 
+#include <cstdio>
+#include <cstdlib>
+#include <string>
+#include <zip.h>
+#include <fmt/core.h>
+#include <feel/feelcore/feel.hpp>
```
prudhomm, line 8: why is this needed here ? you only need <string> that's it

feelpp/feel/feelcore/zip.cpp
```diff
@@ -1,9 +1,3 @@
-#include <cstdio>
-#include <cstdlib>
-#include <string>
-#include <zip.h>
-#include <fmt/core.h>
-#include <feel/feelcore/feel.hpp>
```
prudhomm, line 6: why was this moved ?it is only needed for the implementation not the interface

feelpp/feel/feelcore/zip.cpp
```diff
@@ -39,49 +34,48 @@ bool extractZipFile( const std::string& zipFilePath, const std::string& extracti
         }
 
         fs::path extractionPath = fs::path(extractionDir) / fs::path( entryName );
-        VLOG(2) << fmt::format( "Extracting {} to {}", entryName, extractionPath.string() ) << std::endl;
-        VLOG(2) << fmt::format( "has {} a filename : {}", entryName, fs::path( entryName ).has_filename() ) << std::endl;
-        if ( !fs::path( entryName ).has_filename() || fs::path( entryName ) == fs::path(".") )
+        std::cout << fmt::format( "Extracting {} to {}", entryName, extractionPath.string() ) << std::endl;
```
prudhomm, line 37: this should go in LOG(INFO) or VLOG(1) or VLOG(2)

PR #2437, opened 2026-03-27
resolves 2432: add row support to blockform2

### LINKED ISSUE #2432: add row support to blockform2
to impose Dirichlet conditions strongly it is useful to introduce a new keyword, eg on the first block row `0_c`:

```cpp
a.row( 0_c ) += on( _range = markedfaces( mesh, dirichletMarker ),
                        _rhs = rhs( 0_c ),
                        _element = uh,
                        _expr = cst( potential ),
                        _type = "elimination" );
```

### LINKED ISSUE #2433: support basis functions in dsel
add support for 

```c++
auto u = trial(Xh);
auto v = test(Xh);
auto a = form2(_test=Xh,_trial=Xh);
a = integrate( _range=elements(support(Xh)), _expr=u*v);
```

### LINKED ISSUE #2434: improve form and blockform interface
unify interface and allow `l1Norm(), sum(), .... to avoid going through the algebraic representation

### LINKED ISSUE #2435: modernize static condensation and support 2x2 static condensation
port static condensation to c++20/23 (use concept)
add support for 2x2 blocks in case the unknowns to be eliminated is internal to the element

### LINKED ISSUE #2447: Support deferred Dirichlet conditions
I think it is about time that we implement deferred Dirichlet conditions:
- it is mandatory for static condensation (in hdg, they were imposed weakly, never strongly due to this missing feature)
- if we have multiple on() we don't want to apply `zeroRows()` / `zeroRowsColumns`  multiple times, we want to accumulate the constraints and apply them once for all, limiting the communications to a bare minimum
- finally we want to access the matrices with and without constraints. The latter is critical in many preconditionning strategies

_Originally posted by @prudhomm in https://github.com/feelpp/feelpp/issues/2437#issuecomment-4183114436_

### LINKED ISSUE #2459: Enable PETSc HPDDM auxiliary Neumann matrix support with deferred Dirichlet constraints
Pass the unconstrained operator as HPDDM’s auxiliary Neumann matrix when solving constrained systems produced by deferred Dirichlet conditions, see #2447.
This wires `_auxiliary_matrix` from the form/backend layer down to `PCHPDDMSetAuxiliaryMat()`, keeps HPDDM options applied after auxiliary setup, and updates the HPDDM tests to cover both the plumbing and the deferred-Dirichlet solve path in sequential and MPI runs.

Unfortunately ubuntu and debian do not come with HPDDM support in PETSc, so the work in #2458 is necessary to test the code

/cc @prj- @vincentchabannes

prudhomm:

-----
- closes #2432 
- closes #2433 
- closes #2434 
- closes #2435 
- closes #2447 

closing #2459 has been deferred to the new Problem API to make local neumann matrices auxiliary matrices seamless for the end user.


feelpp/feel/feelalg/matrixpetsc.cpp
```diff
@@ -1767,9 +1767,21 @@ MatrixPetsc<T>::zeroRows( std::vector<int> const& rows,
                           Context const& on_context,
                           value_type value_on_diagonal )
 {
-    // the matrix needs to be closed for this to work
-    if ( !this->closed() )
+    PetscBool assembled = PETSC_FALSE;
+    int ierr = MatAssembled( M_mat, &assembled );
```
vincentchabannes: @prudhomm, this function is not collective (not always the same for all processes). Can generate a deadlock, I think. Some local operations should call collectively to mat->setIsClosed(false)

prudhomm: thanks for the feedback, I will check that

prudhomm: see the correction in [4637241](https://github.com/feelpp/feelpp/pull/2437/commits/4637241d9058af9f9af91313d491220036af8e34)[4637241](https://github.com/feelpp/feelpp/pull/2437/commits/4637241d9058af9f9af91313d491220036af8e34)

ok for you ?

prudhomm:
@vincentchabannes I think it is about time that we implement deferred Dirichlet conditions:
- it is mandatory for static condensation (in hdg, they were imposed weakly, never strongly due to this missing feature)
- if we have multiple on() we don't want to apply `zeroRows()` / `zeroRowsColumns`  multiple times, we want to accumulate the constraints and apply them once for all, limiting the communications to a bare minimum
- finally we want to access the matrices with and without constraints. The latter is critical in many preconditionning strategies
I am working on that solution.

vincentchabannes:
@prudhomm It's exactly what I did in toolboxes (only one call of zeroRow for example). I have also multiple passes, first on Elements, then Faces, then Edges and finally Points. This ordering is important.
For me, only the last point is missing.

feelpp/feel/feelalg/matrixpetsc.cpp
```diff
@@ -757,6 +756,21 @@ void MatrixPetsc<T>::close ()  const
...
+template <typename T>
+inline
+void MatrixPetsc<T>::closeIfNeeded() const
+{
+    if ( !this->isInitialized() )
+        return;
+
+    int needsClose = super::closed() ? 0 : 1;
+    if ( this->comm().size() > 1 )
+        mpi::all_reduce( this->comm(), mpi::inplace( needsClose ), mpi::maximum<int>() );
```
vincentchabannes: @prudhomm You shouldn't call `all_reduce` every time you check whether it's closed or not.
The previous logic worked fine; you just need to place the `setClose(false)` calls in the right places. And if the user manually calls `setValues`, `addMatrix`, etc., they are responsible for closing it or marking it as unclosed.

feelpp/feel/feelalg/matrixpetsc.cpp
```diff
@@ -853,6 +867,7 @@ void MatrixPetsc<T>::set ( const size_type i,
     ierr = MatSetValues( M_mat, 1, &i_val, 1, &j_val,
                          &petsc_value, INSERT_VALUES );
     CHKERRABORT( this->comm(),ierr );
+    this->setIsClosed( false );
```
vincentchabannes: This method can be used intensively, so you should avoid adding a function call

feelpp/feel/feelalg/matrixpetsc.cpp
```diff
@@ -873,6 +888,7 @@ void MatrixPetsc<T>::add ( const size_type i,
     ierr = MatSetValues( M_mat, 1, &i_val, 1, &j_val,
                          &petsc_value, ADD_VALUES );
     CHKERRABORT( this->comm(),ierr );
+    this->setIsClosed( false );
```
vincentchabannes: idem

prudhomm:
@vincentchabannes ok with this version?

feelpp/feel/feelalg/backend.hpp
```diff
@@ -1020,6 +1020,7 @@ class FEELPP_EXPORT Backend : public BackendBase, public std::enable_shared_from
     {
         auto args = NA::make_arguments( std::forward<Ts>(v)... );
         sparse_matrix_ptrtype matrix = args.get(_matrix);
+        sparse_matrix_ptrtype auxiliaryMatrix = args.get_else( _auxiliary_matrix, sparse_matrix_ptrtype{} );
```
vincentchabannes: Only one matrix is too restrictive; for example, AMS requires several matrices. A dictionary (key, matrix) will be more adapted. After, it will be better if theses data are given directly in the preconditioner, and don't make this function any more complex.

prudhomm: do we need to change that now or later in your maxwell branch ?

feelpp/feel/feelalg/matrixpetsc.cpp
```diff
@@ -2799,6 +2814,7 @@ void MatrixPetscMPI<T>::set( const size_type i,
                               &petsc_value, INSERT_VALUES );
 
     CHKERRABORT( this->comm(),ierr );
+    this->setIsClosed( false );
```
vincentchabannes: @prudhomm for me, it's not good to call this mutator here. In //, some process can have no element, but the matrix is still defined. So some process can have the wrong closed properties, and a deadlock will appear. The strategy currently used is to put a closed value globally. When we create bilinearform, lineaform,... we set the closed value before the loop of range mesh elements.

feelpp/feel/feelalg/solvernonlinearpetsc.cpp
```diff
@@ -437,13 +467,17 @@ extern "C"
         int size;
         VecGetSize( x,&size );
 
-        double *xa;
-        VecGetArray( x, &xa );
+        const PetscScalar *xa;
+        VecGetArrayRead( x, &xa );
+        Eigen::Matrix<double, Eigen::Dynamic, 1> x_copy( size );
```
vincentchabannes: Why create a copy?

feelpp/feel/feelalg/backend.hpp
vincentchabannes: No, I attach this directly in the preconditioner object. I also wanted to say that this argument should really be placed in the preconditioner function. That would also make it easier to maintain. Otherwise, we'd have to redo this for each solve, nlSolve, ...

feelpp/feel/feelvf/bilinearformbase.hpp
```diff
@@ -376,12 +424,450 @@ class BilinearFormBase : public CommObject
...
+        {
+            return std::pair{ state.constrainedMatrix, state.constrainedVector };
+        }
+
+        this->get();
+        M_matrix->closeIfNeeded();
+        if ( !rhsVector->closed() )
+            rhsVector->close();
+
+        auto constrainedMatrix = M_matrix->clone();
```
vincentchabannes: @prudhomm, duplicating the matrix can be quite problematic for large cases (memory issue). I think we don't want to create a new matrix, but only keep in memory the values modified by the dof elimination. Just apply a setLocalValues with these entries for create the unconstraint matrix. I may not have fully understood the process

feelpp/feel/feelalg/matrixpetsc.cpp
prudhomm: done

prudhomm:
I am adding a policy for constraints handling
- In-place Dirichlet elimination avoids cloning the complete matrix by default.
- The old two-matrix behavior remains available with _dirichlet_inplace=false
- We store now the RHS correction, essentially -A_FD g_D, rather than every eliminated Dirichlet column. This is sufficient while the operator and constraints remain unchanged; changing them requires reassembly.
- Repeated constraints, ordering/precedence, multiple RHS vectors, elimination variants, block forms, scalar forms, and static condensation are covered.

prudhomm:
@vincentchabannes I have finally remove the systematic cloning of the constrained system from the unconstrained one and have a more powerful API now.

vincentchabannes (approved):

PR #2494, opened 2026-06-18
Feature bvh mpi

vincentchabannes:

-----

1. Full refactoring of BVH intersect algorithm in parallel MPI env. Use TLAS/BLAS approach.

   Implementation keys:
     * Asynchronous MPI communications (ISend,IRecv,IAllreduce,IProbe)
     * Send/recv by batch (max size fixed)
     * Apply MPI transaction in loop event by pulse (min time interval)

   Optimization to investigate:
    * Improve partition prediction:  split the bounding box into a few sub-bounding boxes 
    * Several parameters have been fixed, try to give rules that give better performance (depending on network latency, mesh size, number of partitions, …)

2. The 2D implementation is now based on the same implementation as 3D.
3. NOTE: The context option`BvhIntersectContext::all`. is not correctly implemented in a parallel MPI environment (the result can be incomplete).  The other contexts, `anyHint` and `closest`, are fully operational.


feelpp/feel/feelmesh/bvh.hpp
```diff
@@ -455,36 +417,816 @@ class BVH_ThirdParty : public BVH<MeshEntityType>
...
+                    while (!pending_remote_compute.empty() && chunk_count < CHUNK_SIZE)
+                    {
+                        auto task = pending_remote_compute.front();
+                        pending_remote_compute.pop();
+
+                        auto local_res = this->intersectSequentialImpl<useRobustTraversal,isAnyHint>(task.req.ray, tolerance);
+                        if (!local_res.empty())
+                            pending_outgoing_resps[task.source_rank].push_back({task.req.ray_id, local_res.front()});
+                        else
+                            pending_outgoing_resps[task.source_rank].push_back({task.req.ray_id, rayintersection_result_type(-1, -1, std::numeric_limits<double>::max())});
```
prudhomm, line 962: is that intended ? only local_res.front()  is sent

feelpp/feel/feelmesh/bvh.hpp
```diff
@@ -455,36 +417,816 @@ class BVH_ThirdParty : public BVH<MeshEntityType>
...
 private:
+    std::vector<TlasHit> getHitPartitions(ray_type const& ray) const override
+        {
+            auto intersect_bbox = [](const auto& ray, const auto& bbox) {
+                                      // Initialize with infinity
+                                      value_type t_enter = -std::numeric_limits<value_type>::infinity();
+                                      value_type t_exit  =  std::numeric_limits<value_type>::infinity();
+
+                                      for (int i = 0; i < 3; ++i) {
+                                          auto invD = 1.0 / ray.dir[i];
```
prudhomm: it is dangerous by ray.dir[i] no ? if this is 0 or very small then we get very large number or NaN.

prudhomm (approved):

feelpp/feel/feelmesh/bvh.hpp
vincentchabannes: The “get all intersection” mode is not fully operational in //. This requires changing the communication data structures (which is more resource-intensive). The ‘anyHit’ and “closest” modes are operational.

I will put a note (or maybe also a warning in the log)

feelpp/feel/feelmesh/bvh.hpp
vincentchabannes: @prudhomm Good point! Actually, dividing by 0 gives infinity, and infinity times a number other than 0 is okay. However, there’s a case (very rare) where you can have 0 times infinity, and that’s when it doesn’t work anymore (NaN). I will fix that by adding an epsilon in that case

PR #2514, opened 2026-09-13
reuse function spaces, support static field in ensight, reduce mesh memory usage

### LINKED ISSUE #2512: feat(feeldiscr): reuse function spaces through a mesh-aware manager
Repeated calls to Pch, Pdh, Pchv, Pdhv, Pchm, and similar factories currently rebuild equivalent function spaces and DOF tables, increasing memory use and construction time.

Introduce a mesh-aware function-space manager so equivalent whole-mesh requests reuse an existing space. Provide controls for reuse, forced rebuilding, bypassing, capacity, clearing, and statistics.

Main difficulties:

- identifying a mesh and invalidating entries after mesh changes;
- including all relevant construction options in the repository key;
- keeping repository decisions consistent across MPI ranks;
- avoiding mesh/function-space ownership cycles;
- defining identities for partial mesh supports.

Start with whole-mesh spaces and address range-based spaces later.

### LINKED ISSUE #2513: feat(feelfilters): add static EnSight fields and fix transient export
Support immutable exporter-level static fields alongside time-step fields, reusing P1/P2/P0 spaces through Pch/Pdh. Fix metadata, tensor ordering and packed indexing, with MPI and ParaView regression tests.

 NFS is causing issues with multiple ranks writes, add root level strategy to have only one rank do that. Packed-reader compatibility and mesh changes require explicit constraints; we need to check on lustre filesystem .

Follow-up to #297; related to #2512.

### LINKED ISSUE #2515: perf(feelmesh): bulk face cleanup
Replace repeated face erasure during mesh finalization with bulk cleanup to avoid effectively quadratic behavior on large partitioned meshes.

### LINKED ISSUE #2516: Reduce partition mesh metadata memory usage
Use compact partition metadata and marker fragmentation storage to reduce memory consumption, with regression coverage for mesh metadata and face cleanup.

prudhomm:

-----

- refs #2512
- closes #2513
- closes #2516 
- closes #2515

### DIFF
--- a/benchmarks/feelpp/CMakeLists.txt
+++ b/benchmarks/feelpp/CMakeLists.txt
@@ -23,6 +23,8 @@
 #
 add_custom_target( benchs )
 
+add_subdirectory( functionspace_manager )
+
 feelpp_add_application( env SRCS 01-env.cpp LINK_LIBRARIES  benchmark EXEC B_ENV )
 feelpp_add_application( mesh SRCS 02-mesh.cpp LINK_LIBRARIES  benchmark EXEC B_MESH )
 feelpp_add_application( space SRCS 03-space.cpp LINK_LIBRARIES  benchmark EXEC B_SPACE )
@@ -31,4 +33,5 @@ feelpp_add_application( gmc SRCS gmc.cpp LINK_LIBRARIES  benchmark EXEC B_GMC )
 if ( FEELPP_ENABLE_SPECX )
     feelpp_add_application( specx SRCS 99-specx.cpp LINK_LIBRARIES Feelpp::feelpp_specx )
 endif()
-add_dependencies( benchs ${B_ENV} ${B_MESH} ${B_SPACE} ${B_INTEGRATE} ${B_GMC})
+add_dependencies( benchs ${B_ENV} ${B_MESH} ${B_SPACE} ${B_INTEGRATE} ${B_GMC}
+                  ${B_FSPACE_MANAGER_BASELINE} ${B_FSPACE_MANAGER})
--- a/benchmarks/feelpp/functionspace_manager/CMakeLists.txt
+++ b/benchmarks/feelpp/functionspace_manager/CMakeLists.txt
@@ -1,0 +1,19 @@
+# SPDX-FileContributor: Christophe Prud'homme <christophe.prudhomme@feelpp.org>
+#
+# SPDX-FileCopyrightText: 2026 University of Strasbourg
+#
+# SPDX-License-Identifier: LGPL-3.0-or-later
+
+feelpp_add_application(
+  functionspace_manager_baseline
+  SRCS functionspace_manager_baseline.cpp
+  LINK_LIBRARIES benchmark
+  EXEC B_FSPACE_MANAGER_BASELINE
+)
+
+feelpp_add_application(
+  functionspace_manager
+  SRCS functionspace_manager.cpp
+  LINK_LIBRARIES benchmark
+  EXEC B_FSPACE_MANAGER
+)
--- a/benchmarks/feelpp/functionspace_manager/functionspace_manager.cpp
+++ b/benchmarks/feelpp/functionspace_manager/functionspace_manager.cpp
@@ -1,0 +1,172 @@
+/* -*- mode: c++; coding: utf-8; tab-width: 4; indent-tabs-mode: nil; c-basic-offset: 4 -*-
+
+   SPDX-FileContributor: Christophe Prud'homme <christophe.prudhomme@feelpp.org>
+
+   SPDX-FileCopyrightText: 2026 University of Strasbourg
+
+   SPDX-License-Identifier: LGPL-3.0-or-later
+*/
+
+#include <cstdint>
+#include <set>
+#include <string>
+#include <vector>
+
+#include <benchmark/benchmark.h>
+
+#include <feel/feeldiscr/functionspacebuildinstrumentation.hpp>
+#include <feel/feeldiscr/functionspacemanager.hpp>
+#include <feel/feeldiscr/pch.hpp>
+#include <feel/feeldiscr/pdh.hpp>
+#include <feel/feelfilters/unitcube.hpp>
+#include <feel/feelfilters/unitsquare.hpp>
+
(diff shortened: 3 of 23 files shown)


prudhomm:
@vincentchabannes there were two issues with ensight gold tensor handling
1. Nonsymmetric tensors were transposed on export. Component packing used column-major indexing. It now writes EnSight’s row-major order: `xx, xy, xz, yx, yy, yz, zx, zy, zz`, consistently for nodal, element and cached mappings. See [exporterensightgold_impl.hpp (line 1773)](/nvme0/prudhomm/Devel/feelpp.quickfix/feelpp-2-fspace-repository/feelpp/feel/feelfilters/exporterensightgold_impl.hpp:1773).
2. Tensor expressions overwrote their first column. The conversion loop evaluated each (row, column) component but always assigned it to [row][0]. It now fills [row][column], with MPI synchronization enabled. See [exportfieldset.hpp (line 699)](/nvme0/prudhomm/Devel/feelpp.quickfix/feelpp-2-fspace-repository/feelpp/feel/feeldiscr/exportfieldset.hpp:699).

exportfieldset is an extraction of timeset.hpp to avoid duplication for time independent and time dependent handling, it is shared. timeset.hpp has been substantially reduced and use exportfieldset

prudhomm:
the functionmanager still needs to be applied to pchv and pchm/pchms

prudhomm:
@vincentchabannes only one test failed in both spack and ubuntu consistently because I changed the policy to use reuse by default now

I suggest that you review this PR, it would not close #2512 since mesh ranges still need to be supported

prudhomm:
for  partial mesh support, we need range-based factories and they remain  uncached for now.

in that case the manager needs an immutable, value-based element-range signature in both its cache key and MPI consistency . This must distinguish different supports while allowing independently-created equivalent ranges to reuse.

```cpp
// Target API/behaviour
auto Xh1 = Pch<1>( _mesh=mesh, _range=markedelements( mesh, "Omega1" ) );
auto Xh2 = Pch<1>( _mesh=mesh, _range=markedelements( mesh, "Omega1" ) );
auto Yh  = Pch<1>( _mesh=mesh, _range=markedelements( mesh, "Omega2" ) );

CHECK( Xh1 == Xh2 ); // same element support: reuse
CHECK( Xh1 != Yh );  // different element support: distinct space of course
```

Initially we should cover the usual spaces `Pch`, `Pdh`, `Pchv`, and `Pdhv`, with serial and MPI validation.

feelpp/feel/feelmesh/compactmarkerfragmentation.hpp
```diff
@@ -0,0 +1,139 @@
...
+    boost::mpi::reduce( comm, &local, 1, &global, MergeMarkerBuffers<value_type>{}, 0 );
+    std::vector<value_type>().swap( local );
+    std::uint64_t length = global.size();
+    boost::mpi::broadcast( comm, length, 0 );
+    global.resize( length );
+    constexpr std::size_t chunkSize = 1024 * 1024;
+    for ( std::size_t offset = 0; offset < global.size(); offset += chunkSize )
+    {
+        int chunk = std::min( chunkSize, global.size() - offset );
+        boost::mpi::broadcast( comm, global.data() + offset, chunk, 0 );
```
vincentchabannes, line 124: Why not broadcast all data?

feelpp/feel/feelfilters/detail/fileindex.cpp
```diff
@@ -23,156 +23,212 @@
...
+            if ( flag != 0 )
+                checkIndexIo( MPI_ERR_IO, this->worldComm() );
+            LOG( INFO ) << "Done reading FILE_INDEX";
         }
     }
 
-    mpi::broadcast( this->worldComm(), M_fileblocks, this->worldComm().masterRank() );
-    mpi::broadcast( this->worldComm(), M_nextFreePosFile, this->worldComm().masterRank() );
+    mpi::broadcast( this->worldComm().comm(), M_fileblocks, this->worldComm().masterRank() );
+    mpi::broadcast( this->worldComm().comm(), M_nextFreePosFile, this->worldComm().masterRank() );
```
vincentchabannes, line 170: why necessary?

feelpp/feel/feelfilters/detail/fileindex.cpp
```diff
@@ -23,156 +23,212 @@
...
     MPI_Status status;
 
-    LOG(INFO) << "Start writing FILE_INDEX";
+    LOG( INFO ) << "Start writing FILE_INDEX";
+    // The next append overwrites this trailer, not the end of the whole file.
+    M_nextFreePosFile = offset;
 
     /* only process 0 writes data */
-    if ( this->worldComm().isMasterRank() )
+    if ( this->worldComm().comm().rank() == this->worldComm().masterRank() )
```
vincentchabannes, line 186: why isMasterRank is not good?

vincentchabannes (approved):
Great job!

feelpp/feel/feelmesh/compactmarkerfragmentation.hpp
prudhomm: The fragmentation table is large in some production meshes. For example, the Chicago 6 km LOD1 mesh has more than 500K physical groups (including more than 100K buildings). The representation stores records as [length, marker-id...]; even one ID per group exceeds one million primitive values. (This is the last change I made to make Chicago work reasonably.)
We therefore broadcast the raw data in chunks after broadcasting the total length. This avoids relying on one very large serialized Boost.MPI vector transfer and keeps each primitive MPI call with  fixed count/size. There is a chunk regression test that covers this.

feelpp/feel/feelfilters/detail/fileindex.cpp
prudhomm: good catch, I was testing different strategies on gaya for the NFS I/O at the time and didn't clean that.

prudhomm: `this->worldComm().comm()` sends back the `localComm()` whereas `this->worldComm()` sends back the `globalComm()`

